PR #2, opened 2015-06-26
PG-206 Localized Character Names with ParaTExt BiblicalTerms files

grant-dennison:
Added new dev tool "TermTranslator" which takes a BiblicalTerms file from the DevTools/Resources folder and generates a tmx file in DistFiles/localization.


DevTools/Program.cs
```diff
@@ -6,8 +6,9 @@ static void Main(string[] args)
 		{
 //			CharacterListProcessing.Process();
 //			BiblicalTerms.Processor.Process();
-			FCBH.Processor.Process();
+//			FCBH.Processor.Process();
 //			VerseBridgeHelper.RemoveAllVerseBridges();
+            TermTranslator.Processor.Process();
```
andrew-polk: I think everything in Glyssen currently uses tabs.  Let's change your VS settings to tabs.

DevTools/TermTranslator/tmxFormat.cs
```diff
@@ -0,0 +1,144 @@
+﻿using System;
```
andrew-polk: Go ahead and rename this file to TmxFormat for consistency and conformance to standards.

PR #3, opened 2015-06-26
PG-253 Tooltip now shows on hover over any part of context

grant-dennison:
When the mouseover event is triggered, up to ten parents are checked above the target to see if the element is a context.

Tooltip is only shown when no tooltip is present.

Tooltip is only hidden when an area that isn't a context is entered.

### DIFF
--- a/Glyssen/Controls/ScriptBlocksViewer.cs
+++ b/Glyssen/Controls/ScriptBlocksViewer.cs
@@ -163,45 +163,54 @@ namespace Glyssen.Controls
 				e.Value = m_getDelivery(block);
 			}
 		}
+
+		private void HideToolTip()
+		{
+			if(m_toolTip != null)
+			{
+				m_toolTip.Hide(this);
+				m_toolTip.Dispose();
+				m_toolTip = null;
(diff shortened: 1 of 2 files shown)


Glyssen/Controls/ScriptBlocksViewer.cs
```diff
@@ -163,45 +163,54 @@ private void HandleDataGridViewBlocksCellValueNeeded(object sender, DataGridView
...
-			var divElement = geckoElement as GeckoDivElement;
-			if (divElement == null)
-				return;
+			var checkElement = geckoElement as GeckoHtmlElement;
 
-			if (divElement.Parent.ClassName == BlockNavigatorViewModel.kCssClassContext)
+			for (int i = 0; i < 10; i++)
 			{
-				m_toolTip = new ToolTip {IsBalloon = true};
+				if(checkElement == null)
```
andrew-polk: I know I'm nitpicking here, but "normal" C# style is no space after method names but spaces after for, if, etc.
Please fix this one and the one on else if below.

PR #5, opened 2015-06-30
PG-234 Initial construction of Voice Actor info dialog

grant-dennison:
Constructed shell of Voice Actor Information window based on design document.
I was unable to implement a move to next field on combobox item selection.

Todo: implement buttons (save/next)

### DIFF
--- a/Glyssen/Controls/DataGridViewOverrideEnter.cs
+++ b/Glyssen/Controls/DataGridViewOverrideEnter.cs
@@ -1,14 +1,11 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
-using System.Windows.Forms;
+﻿using System.Windows.Forms;
 using SIL.Windows.Forms.Widgets.BetterGrid;
 
 namespace Glyssen.Controls
 {
-	//DataGridView with Enter moving to right (instead of down)
+	/// <summary>
+	/// DataGridView with Enter moving to right (instead of down)
+	/// </summary>
 	public class DataGridViewOverrideEnter : BetterGrid
 	{
 		public DataGridViewOverrideEnter()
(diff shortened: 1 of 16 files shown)


Glyssen/Controls/DataGridViewOverrideEnter.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Forms;
+using SIL.Windows.Forms.Widgets.BetterGrid;
+
+namespace Glyssen.Controls
+{
+	//DataGridView with Enter moving to right (instead of down)
+	class DataGridViewOverrideEnter : BetterGrid
```
andrew-polk: Let's go ahead and make this public unless there is a reason not to.

Glyssen/Dialogs/VoiceActorInformationDlg.cs
```diff
@@ -16,5 +18,69 @@ public VoiceActorInformationDlg()
...
 			InitializeComponent();
 		}
+
+		private void RemoveSelectedRows(bool confirmWithUser)
+		{
+			bool deleteConfirmed = !confirmWithUser;
+
+			if (confirmWithUser)
+			{
+				deleteConfirmed = MessageBox.Show("Are you sure you want to delete these rows?", "Confirm", MessageBoxButtons.YesNo) == System.Windows.Forms.DialogResult.Yes;
```
andrew-polk: Any message which gets displayed to the user needs to be wrapped in LocalizationManager.GetString(...).  Better yet, see if you can make use of the dialog in libpalaso.  I think it is called something like recycle*dialog or delete*dialog.

Glyssen/Dialogs/VoiceActorInformationDlg.cs
```diff
@@ -16,5 +18,69 @@ public VoiceActorInformationDlg()
...
+			}
+
+			if (deleteConfirmed)
+			{
+				for (int i = m_dataGrid.SelectedRows.Count - 1; i >= 0; i--)
+				{
+					m_dataGrid.Rows.Remove(m_dataGrid.SelectedRows[i]);
+				}
+
+				if(m_dataGrid.RowCount <= 1)
```
andrew-polk: Space after if.  You can probably change your settings in VS if it isn't doing this by default.

PR #11, opened 2015-07-10
PG-263 Populate "Hours" field in VA Assign

grant-dennison:
Current implementation utilizing calculation based on keystrokes spoken per hour.


Glyssen/Dialogs/VoiceActorAssignmentDlg.cs
```diff
@@ -39,6 +42,28 @@ public VoiceActorAssignmentDlg(Project project)
...
+
+				foreach (var group in characterGroups)
+				{
+					int keyStrokes = 0;
+					foreach (var characterId in group.CharacterIds)
+					{
+						if (keyStrokesByCharacterId.ContainsKey(characterId))
+							keyStrokes += keyStrokesByCharacterId[characterId];
+					}
+					group.EstimatedHours = (keyStrokes/kKeyStrokesPerHour).ToString("0.00");
```
andrew-polk: Let's see if EstimatedHours can be a double to ensure proper sorting in the grid

PR #12, opened 2015-07-14
PopulateEstimatedHours only needs to be run the first time

grant-dennison:


Glyssen/Dialogs/VoiceActorAssignmentDlg.cs
```diff
@@ -39,9 +39,9 @@ public VoiceActorAssignmentDlg(Project project)
 
 				foreach (KeyValuePair<int, CharacterGroup> pair in charGroupTemplate.CharacterGroups)
 					characterGroups.Add(pair.Value);
-			}
 
-			m_project.CharacterGroupList.PopulateEstimatedHours(m_project.IncludedBooks);
+				m_project.CharacterGroupList.PopulateEstimatedHours(m_project.IncludedBooks);
```
andrew-polk: Won't this make it so that the values won't be populated when returning to the grid ?

grant-dennison: Estimated hours are stored in characterGroups.xml along with other group information. This makes sense because we are also storing characterIds. When I tested it, it loaded correctly.

andrew-polk: Ah.  Didn't realize we were persisting that.  I guess in the case I saw in which it didn't populate, my character group file must have been created before the code was implemented.

Storing the characterIds is different because that is actually part of the definition of a group, but I don't see any reason not to store the estimated hours.

PR #30, opened 2015-07-27
Save Status

grant-dennison:


Glyssen/Controls/VoiceActorInformationGrid.cs
```diff
@@ -53,6 +54,11 @@ public void Initialize(Project project)
 
 		public void SaveVoiceActorInformation()
 		{
+			if (Save != null)
```
andrew-polk: Probably better to call the event after the save happens.  If a problem happens during the actual save, we don't want to call the event.

PR #38, opened 2015-07-30
PG-308 Expand Character Groups

grant-dennison:


Glyssen/Controls/VoiceActorInformationGrid.cs
```diff
@@ -29,6 +29,8 @@ public VoiceActorInformationGrid()
 		{
 			InitializeComponent();
 
+			m_dataGrid.AutoSizeRowsMode = DataGridViewAutoSizeRowsMode.None;
```
grant-dennison: This keeps the VA Info rows from being bigger than VA Assign rows.

PR #48, opened 2015-10-27
Get LDML from bundle if it exists (PG-259)

andrew-polk:


Glyssen/Project.cs
```diff
@@ -86,9 +85,14 @@ private Project(GlyssenDblTextMetadata metadata, string recordingProjectName = n
 		}
 
 		public Project(GlyssenBundle bundle, string recordingProjectName = null, Project projectBeingUpdated = null) :
-			this(bundle.Metadata, recordingProjectName)
+			this(bundle.Metadata, recordingProjectName, false, bundle.WritingSystemDefinition)
 		{
 			Directory.CreateDirectory(ProjectFolder);
+			if (bundle.WritingSystemDefinition != null && bundle.WritingSystemDefinition.QuotationMarks != null && bundle.WritingSystemDefinition.QuotationMarks.Any())
```
tombogle: I think it would be better to move ConvertContinuersToParatextAssumptions into Project because it's not obvious that the WritingSystemDefinition in the bundle is the same object as the one pointed to by m_wsDefnition here.

Glyssen/Project.cs
```diff
@@ -339,10 +348,58 @@ public Project UpdateProjectFromBundleData(GlyssenBundle bundle)
...
+			{
+			}
+			// Copy if source has more detail
+			else
+			{
+				bool copy = false;
+				var sourceLevelsCount = WritingSystem.QuotationMarks.Count;
+				var targetLevelsCount = targetWs.QuotationMarks.Count;
+
+				if (sourceLevelsCount >= targetLevelsCount)
```
tombogle: I think this should be strictly greater than. (Though there should be no harm in copying it if it's identical, so it probably doesn't matter in practice.)

GlyssenTests/ProjectTests.cs
```diff
@@ -1,10 +1,205 @@
...
+using GlyssenTests.Bundle;
 using NUnit.Framework;
+using SIL.IO;
+using SIL.ObjectModel;
+using SIL.WritingSystems;
 
 namespace GlyssenTests
 {
-	internal class ProjectTests
+	class ProjectTests
```
tombogle: Running all these tests together as a batch using Resharper in VS causes a series of "The process cannot access the file" messages, and the CreateFromBundle tests hang.

Glyssen/Project.cs
andrew-polk: Done

Glyssen/Project.cs
andrew-polk: Done

PR #57, opened 2015-11-11
Convert IsCameo property to SpecialRole and update VoiceActorInfoDlg PG-452 and Add Narration Options Dlg PG-475

rick-conrad-sil:
I guess we want to wait on this one until after Sprint 13 now...


Glyssen/Bundle/GlyssenDblTextMetadata.cs
```diff
@@ -263,6 +268,27 @@ public QuoteSystemStatus QuoteSystemStatus
...
+	public class ProjectSettings
+	{
+		public const int kNumberOfNarratorsNotSet = 0;
+
+		/// <summary>
+		/// Number of narrators set by user
+		/// 0 if not set by user
+		/// </summary>
+		/// 
+		[XmlElement("numberofnarrators")]
```
tombogle: Typically, we use attributes for simple data types and elements for more complex stuff. Did you consider using attributes? If so, why did you choose to use elements? Because there might end up being a lot of settings, I could maybe see the benefit in using elements (to make the XML easier to read). Another option would be to have sub-elements under ProjectSettings to group the different settings. For now, we would just have "NarratorSettings", but eventually there would be others. Then each of those elements could use attributes for the simple int/bool/enum types.

Glyssen/Bundle/GlyssenDblTextMetadata.cs
```diff
@@ -84,6 +85,9 @@ public override GlyssenDblMetadataLanguage Language
 		[XmlElement("projectStatus")]
 		public ProjectStatus ProjectStatus = new ProjectStatus();
 
+		[XmlElement("projectsettings")]
```
tombogle: I realize that the immediately preceding element does not conform to this, but I think we typically use an initial capital letter for elements and an initial lowercase letter for attributes.

Glyssen/Dialogs/NarrationOptionsDlg.cs
```diff
@@ -0,0 +1,91 @@
...
+			else
+				m_project.ProjectSettings.NarratorGenders = NarratorGenders.MaleOnly;
+
+			m_project.Save();
+
+			DialogResult = DialogResult.OK;
+			Close();
+		}
+
+		private void m_btnCancel_Clicked(object sender, EventArgs e)
```
tombogle: You can accomplish this via the Designer by setting the DialogResult property of the Cancel button. (Also, the dialog's CancelButton property should be set to this button.

Glyssen/Dialogs/NarrationOptionsDlg.cs
```diff
@@ -0,0 +1,91 @@
...
+			if(m_rdoFemaleOnly.Checked)
+				m_project.ProjectSettings.NarratorGenders = NarratorGenders.FemaleOnly;
+			else if(m_rdoMaleOrFemale.Checked)
+				m_project.ProjectSettings.NarratorGenders = NarratorGenders.MaleOrFemale;
+			else
+				m_project.ProjectSettings.NarratorGenders = NarratorGenders.MaleOnly;
+
+			m_project.Save();
+
+			DialogResult = DialogResult.OK;
```
tombogle: You can accomplish this via the Designer by setting the DialogResult property of the OK button. (Also, the dialog's AcceptButton property should be set to this button.

Glyssen/Dialogs/NarrationOptionsDlg.cs
```diff
@@ -0,0 +1,91 @@
...
+					m_rdoMaleOrFemale.Checked = true;
+					break;
+				default:
+					m_rdoMaleOnly.Checked = true;
+					break;
+
+			}
+		}
+
+		private bool m_validateAndSetNarratorNum()
```
tombogle: Most of this code can be eliminated by using a numeric up/down control instead of a text box. Note: we probably want to set a max value equal to the number of books in the project.

Glyssen/Dialogs/NarrationOptionsDlg.cs
```diff
@@ -0,0 +1,91 @@
...
+		private readonly Project m_project;
+
+		public NarrationOptionsDlg(Project project)
+		{
+			InitializeComponent();
+			m_project = project;
+
+			if (m_project.ProjectSettings.NumberOfNarrators != ProjectSettings.kNumberOfNarratorsNotSet)
+			{
+				m_intNarratorNum = m_project.ProjectSettings.NumberOfNarrators;
```
tombogle: If this has not been set before, I think we want to automatically set it to a valid default value that makes sense. Not sure if it should be 1 (which would mimic the current behavior) or a value calculated based on some heuristic that reflects the books included in the project.

Glyssen/Bundle/GlyssenDblTextMetadata.cs
```diff
@@ -214,6 +218,7 @@ public void CopyGlyssenModifiableSettings(GlyssenDblTextMetadata source)
 			FontFamily = source.FontFamily;
```
tombogle: FontFamily and FontSizeInPoints, and ProjectStatus are also project settings. Should we maybe call this CharacterGroupGenerationPreferences instead?

Glyssen/Bundle/GlyssenDblTextMetadata.cs
rick-conrad-sil: I did attributes at first (not noticing the difference). Andrew told me to change it to elements and described what he wanted, so I worked it up that way

Glyssen/Bundle/GlyssenDblTextMetadata.cs
rick-conrad-sil: I see...looks like attributes are all lowercase and elements are initial lowercase and then camel case.

andrew-polk:
PG-475 is complete.  PG-452 is on hold indefinitely.  Closing this for now but leaving the VoiceActorInfoDlg-SetNarrators-PG452 in place.

PR #60, opened 2016-01-04
Set the Glyssen data directory to be full control for everyone

phillip-hopper:
Also check if the project file is writable.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/60)

### DIFF
--- a/Glyssen/Program.cs
+++ b/Glyssen/Program.cs
@@ -73,6 +73,15 @@ namespace Glyssen
 				if (Directory.Exists(oldPgBaseFolder) && !Directory.Exists(BaseDataFolder))
 					Directory.Move(oldPgBaseFolder, BaseDataFolder);
 
+				if (!Directory.Exists(BaseDataFolder))
+				{
+					// create the directory
+					Directory.CreateDirectory(BaseDataFolder);
+				}
+
+				// PG-433, 07 JAN 2016, PH: Set the permissions so everyone can read and write to this directory
+				DirectoryUtilities.SetFullControl(BaseDataFolder, false);
+
 				DataMigrator.UpgradeToCurrentDataFormatVersion();
 
 				SampleProject.CreateSampleProjectIfNeeded();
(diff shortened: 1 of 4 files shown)


tombogle:
1) I'm wondering whether the file utilities stuff is (or should be) part of libpalaso. Seems like it's general-purpose code that belongs there rather than in Glyssen.
2) I think we should disable parts of the UI for read-only projects. It doesn't make sense to tell them that changes won't be saved but then continue to allow them to work as if all is well. They should be able to export scripts from a read-only project and maybe visit the project properties dialog (which should also show that the project is read-only), but most of the rest of the UI should just be disabled.

andrew-polk:
What about migration of existing installs?  For example, this issue seems to have been causing Mark Scott quite a bit of grief.  We don't want him to have to jump through any hoops to get everything working.

phillip-hopper:
@tombogle: I put the file utility code in Palaso, and disabled most of the UI if the project file is read-only.

phillip-hopper:
@andrew-polk: I set the code to check each time the program starts rather than just when the directory is created in ProgramData.

Glyssen/Program.cs
```diff
@@ -73,6 +73,15 @@ static void Main()
 				if (Directory.Exists(oldPgBaseFolder) && !Directory.Exists(BaseDataFolder))
 					Directory.Move(oldPgBaseFolder, BaseDataFolder);
 
+                if (!Directory.Exists(BaseDataFolder))
```
andrew-polk: It looks like these lines are indented with spaces rather than tabs.

phillip-hopper: Fixed.

tombogle:
I think the changes to MainForm are not right. See method UpdateButtons. I think what you need to do is to ensure that the readonly flag gets passed in to that method correctly. As it is, I believe you will end up enabling some of those buttons in situations where they shouldn't be.

PR #61, opened 2016-01-05
Do not remove the group when the role is removed from a cameo actor

phillip-hopper:
[Review on Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/61)


tombogle:
Changes look good, but it'd be nice to see a passing build. (We really need to figure out what it's going to take to get the build to be more reliable.)

PR #65, opened 2016-01-14
PG-508: Remember voice actor window sizes

phillip-hopper:
Also remember character group grid settings

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/65)


andrew-polk:
Good approach.  Have you considered how/if this affects the dialogs/grids which we already store/restore settings for?

PR #66, opened 2016-01-19
PG-513: Display location of most recent export

phillip-hopper:
Also provide a link to open the directory

Resolves #PG-513

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/66)

### DIFF
--- a/Glyssen/Dialogs/ExportDlg.cs
+++ b/Glyssen/Dialogs/ExportDlg.cs
@@ -159,6 +159,9 @@ namespace Glyssen.Dialogs
 			try
 			{
 				m_projectExporter.GenerateFile(filePath, m_selectedFileType);
+
+				// remember the location
+				m_projectExporter.Project.Status.LastExportLocation = Path.GetDirectoryName(filePath);
 			}
 			catch (Exception ex)
 			{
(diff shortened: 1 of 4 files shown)


Glyssen/Dialogs/ExportDlg.cs
```diff
@@ -159,6 +159,10 @@ private void BtnOk_Click(object sender, EventArgs e)
 			try
 			{
 				m_projectExporter.GenerateFile(filePath, m_selectedFileType);
+
+				// remember the location
+				Settings.Default.LastExportLocation = Path.GetDirectoryName(filePath);
```
andrew-polk: I think this needs to be per project, thus needs to be saved in the project metadata.

andrew-polk:
Also, its just an opinion (and you could check with someone else for theirs), but I think the link should line up horizontally with the text next to the other buttons (rather than the check marks).

Glyssen/Dialogs/ExportDlg.cs
phillip-hopper: Done.

phillip-hopper:
Horizontal alignment updated.

PR #71, opened 2016-02-02
Fix some problems with m_characterGroupGrid

phillip-hopper:
Ignore header row in m_characterGroupGrid_CellContentClick
Fix sorting direction icon in headers
Allow user to click `OK` after deleting the last actor

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/71)

### DIFF
--- a/Glyssen/Dialogs/VoiceActorAssignmentDlg.cs
+++ b/Glyssen/Dialogs/VoiceActorAssignmentDlg.cs
@@ -420,6 +420,9 @@ namespace Glyssen.Dialogs
 
 		private void m_characterGroupGrid_CellContentClick(object sender, DataGridViewCellEventArgs e)
 		{
+			// ignore clicks on the header row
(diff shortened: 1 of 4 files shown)


Glyssen/Dialogs/VoiceActorAssignmentDlg.cs
```diff
@@ -420,6 +420,9 @@ private void m_btnCancelMove_Click(object sender, EventArgs e)
 
 		private void m_characterGroupGrid_CellContentClick(object sender, DataGridViewCellEventArgs e)
 		{
+			// ignore clicks on the header row
+			if (e.RowIndex < 0) return;
```
andrew-polk: Won't this prevent the user from sorting?  I don't think we want to do that unless we have to.

andrew-polk: Phil says no; the user can still sort.

PR #73, opened 2016-02-10
PG-541: Clear form and show progress when opening project

phillip-hopper:
Resolves #PG-541

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/73)

### DIFF
--- a/Glyssen/MainForm.cs
+++ b/Glyssen/MainForm.cs
@@ -133,6 +133,13 @@ namespace Glyssen
 			m_imgCheckBooks.Visible = false;
 			m_imgCheckAssignCharacters.Visible = false;
 			m_imgCheckAssignActors.Visible = false;
+
+			m_lblProjectInfo.Text = string.Empty;
+			m_lblSettingsInfo.Text = string.Empty;
+			m_lblBookSelectionInfo.Text = string.Empty;
+			m_lblPercentAssigned.Text = string.Empty;
+			m_lblActorsAssigned.Text = string.Empty;
+			m_lastExportLocationLink.Text = string.Empty;
 		}
 
 		private void HandleOpenProject_Click(object sender, EventArgs e)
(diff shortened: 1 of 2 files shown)


Glyssen/MainForm.cs
```diff
@@ -459,12 +523,7 @@ private void SetupUiLanguageMenu()
 
 		private void Assign_Click(object sender, EventArgs e)
 		{
-			using (var viewModel = new AssignCharacterViewModel(m_project))
```
andrew-polk: Was this removed on purpose?

Glyssen/MainForm.cs
```diff
@@ -172,24 +180,69 @@ private void MainForm_Load(object sender, EventArgs e)
...
+			m_progressBar.Visible = false;
+			m_tableLayoutPanel.Enabled = true;
+			UpdateDisplayOfProjectInfo();
+		}
+
+		private async void LoadProject(string filePath)
+		{
+			InitializeProgress(3);
+			IncrementProgress();
+			await Task.Delay(300);
```
andrew-polk: Do these calls actually make it take longer to load a project?  If so, would we be better off just having a marquee progress bar which goes until the load is complete?

andrew-polk:
When I ran it, it looks like the label and all the buttons got moved down a little bit on the Main Form.  Now it looks like there is extra space at the top.

Glyssen/MainForm.cs
phillip-hopper: Fixed.

Glyssen/MainForm.cs
phillip-hopper: Removed the delays and made changed to marque progress bar.

phillip-hopper:
Removed most of the top padding.

phillip-hopper:
Removed the progress bar.

Glyssen/MainForm.cs
```diff
@@ -172,6 +187,20 @@ private void MainForm_Load(object sender, EventArgs e)
 				LoadProject(Settings.Default.CurrentProject);		
 		}
 
+		private void InitializeProgress()
+		{
+			ResetUi();
+			m_tableLayoutPanel.Enabled = false;
+			Cursor = Cursors.WaitCursor;
```
andrew-polk: It appears this should be 
Cursor.Current = Cursors.WaitCursor;

The way it is now, it seems the UI still never updates with the correct cursor.

Glyssen/MainForm.cs
```diff
@@ -172,6 +187,20 @@ private void MainForm_Load(object sender, EventArgs e)
...
+		}
+
+		private void CloseProgress()
+		{
+			Cursor = Cursors.Default;
```
andrew-polk: See comment above.

Glyssen/MainForm.cs
phillip-hopper: I made the change, although I didn't notice any difference in behavior.

PR #76, opened 2016-02-11
Issue PG-544: Don't show test results until until quote parser runs

phillip-hopper:
Changed QuoteParser so changes can be tested can be run without updating the project.

Changed BlockNavigatorViewModel so it can show blocks that are not part of a project (output from the Quoteparser changes).

Resolves #PG-544

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/76)

### DIFF
--- a/Glyssen/Quote/QuoteParser.cs
+++ b/Glyssen/Quote/QuoteParser.cs
@@ -42,6 +42,30 @@ namespace Glyssen.Quote
 			projectWorker.ReportProgress(100);
 		}
 
+		public static List<BookScript> TestQuoteSystem(Project project, QuoteSystem altQuoteSystem)
+		{
+			var cvInfo = new CombinedCharacterVerseData(project);
+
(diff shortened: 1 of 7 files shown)


Glyssen/Quote/QuoteParser.cs
```diff
@@ -42,6 +42,28 @@ public static void ParseProject(Project project, BackgroundWorker projectWorker)
 			projectWorker.ReportProgress(100);
 		}
 
+		internal static List<BookScript> TestQuoteSystem(Project project, QuoteSystem altQuoteSystem)
```
andrew-polk: Unless I am missing something, it seems that this test parse starts with the books/blocks in their current state.  We need to parse from a fresh (unparsed) state.  The discussed strategy means writing some kind of Unparse algorithm to reverse engineer the parse back to the original state.

phillip-hopper:
Pull request updated.

PR #78, opened 2016-02-18
Issue PG-171: Make the quote system in the quote parser static

phillip-hopper:
Consider making the quote system in the quote parser a static member to improve performance.

Resolves #PG-171

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/78)

### DIFF
--- a/Glyssen/Quote/QuoteParser.cs
+++ b/Glyssen/Quote/QuoteParser.cs
@@ -32,9 +32,10 @@ namespace Glyssen.Quote
 			int allProjectBlocks = numBlocksPerBook.Values.Sum();
 
 			int completedProjectBlocks = 0;
+			SetQuoteSystem(project.QuoteSystem);
 			Parallel.ForEach(blocksInBook.Keys, book =>
 			{
-				book.Blocks = new QuoteParser(cvInfo, book.BookId, blocksInBook[book], project.QuoteSystem, project.Versification).Parse().ToList();
+				book.Blocks = new QuoteParser(cvInfo, book.BookId, blocksInBook[book], project.Versification).Parse().ToList();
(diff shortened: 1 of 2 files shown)


GlyssenTests/Quote/QuoteParserTests.cs
```diff
@@ -28,6 +27,7 @@ public void Parse_ContainsUserConfirmedBlock_ThrowsInvalidOperationException()
 			block.BlockElements.Add(new ScriptText("He replied, «Isaiah was right when he prophesied about you.»"));
 			block.UserConfirmed = true;
 			var input = new List<Block> { block };
+			QuoteParser.SetQuoteSystem(QuoteSystem.Default);
```
andrew-polk: Is this needed since this default is set in QuoteParser?

Glyssen/Quote/QuoteParser.cs
```diff
@@ -66,11 +68,16 @@ public static List<BookScript> TestQuoteSystem(Project project, QuoteSystem altQ
...
+		{
+			m_quoteSystem = quoteSystem;
+		}
+
 		private readonly ICharacterVerseInfo m_cvInfo;
 		private readonly string m_bookId;
 		private readonly int m_bookNum;
 		private readonly IEnumerable<Block> m_inputBlocks;
-		private readonly QuoteSystem m_quoteSystem;
+		private static QuoteSystem m_quoteSystem = QuoteSystem.Default;
```
andrew-polk: Now that this is static, should be s_quoteSystem

GlyssenTests/Quote/QuoteParserTests.cs
phillip-hopper: Yes, because other tests set the QuoteSystem to non-default values.

PR #120, opened 2016-04-04
PG-651: Changed Voice Actor List dialog to offer both OK and Cancel b…

tombogle:
…uttons when initially displayed via Cast Size Planning dialog. Clicking OK attempts to select Actual Cast Size option (if there are enough actors). Clicking Cancel leaves the pre-existing cast size option selected.

PG-653: Cast size on main form reflects actual number of groups in Project
PG-654: When user generates groups, the new groups are saved (replacing previously generated groups).
Also fixed wording (and the minimum cast size number shown) in the warning message when the number of actors is below the minimum.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/120)

### DIFF
--- a/Glyssen/Character/CharacterGroupList.cs
+++ b/Glyssen/Character/CharacterGroupList.cs
@@ -27,7 +27,7 @@ namespace Glyssen.Character
 		public ObservableList<CharacterGroup> CharacterGroups
 		{
 			get { return m_characterGroups; }
-			set 
+			private set 
 			{
 				m_characterGroups = value;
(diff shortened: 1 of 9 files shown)


phillip-hopper:
:lgtm:

---

Reviewed 9 of 9 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/glyssen/120#-:-KEX-YxTGCG4iBhHC9f0:1417088721)*

PR #121, opened 2016-04-04
PG-408: Remove last voice actor line if name is blank

phillip-hopper:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/121)


tombogle:
I've added a comment to the Jira issue. I think we want to get clarification on the desired behavior. We may need to expand this fix to cover the more troublesome case of a duplicate name.

andrew-polk:
Apparently this pull request got automatically closed when I deleted the BetaReleaseForEMDC branch. It should be recreated (at some point) against master.

PR #203, opened 2016-06-15
PG-741: Prevent extra breaks in vernacular from `poetic` blocks

phillip-hopper:
If a `poetic` block has the same speaker as the previous block, combine the block with the previous one

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/203)

### DIFF
--- a/Glyssen/BookScript.cs
+++ b/Glyssen/BookScript.cs
@@ -8,6 +8,7 @@ using Glyssen.Dialogs;
 using Glyssen.Quote;
 using SIL.ObjectModel;
 using SIL.Scripture;
+using SIL.Unicode;
 using ScrVers = Paratext.ScrVers;
 
 namespace Glyssen
@@ -20,6 +21,7 @@ namespace Glyssen
 		private int m_blockCount;
 		private List<Block> m_blocks;
 		private List<List<Block>> m_unappliedSplitBlocks = new List<List<Block>>();
+		private ScrStylesheetAdapter m_styleSheet;
 
 		public BookScript()
 		{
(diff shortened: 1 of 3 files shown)


andrew-polk:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/glyssen/203#-:-KKKZ-V8iBLnSZ1vVE_z:1417088721)*

Glyssen/BookScript.cs
```diff
@@ -102,36 +103,50 @@ public System.Collections.Generic.IReadOnlyList<Block> GetScriptBlocks(bool join
...
-						}
-						foreach (var blockElement in block.BlockElements.Skip(skip))
-							newBlock.BlockElements.Add(blockElement.Clone());
-						newBlock.UserConfirmed &= block.UserConfirmed;
-						list[list.Count - 1] = newBlock;
+						list[list.Count - 1] = CombineBlockWithPreviousBlock(block, prevBlock);
 						continue;
 					}
 				}
+				else if (block.StyleTag.StartsWith("q")
```
tombogle: Seems like it would be "better" to access the IsPoetic flag from the "TextProperties" of the style from the USFM stylesheet, rather than relying on the convention of starting these markers with the letter "q". Not sure how easily we can fluff up a stylsheet in this context (see USXParser for example).

Glyssen/BookScript.cs
```diff
@@ -102,36 +103,50 @@ public System.Collections.Generic.IReadOnlyList<Block> GetScriptBlocks(bool join
...
-			list.Add(m_blocks[0]);
-			for (int i = 1; i < m_blockCount; i++)
+			var list = new List<Block>(m_blockCount) {m_blocks[0]};
+
+			for (var i = 1; i < m_blockCount; i++)
 			{
 				var block = m_blocks[i];
+				var prevBlock = list.Last();
+
 				if (!block.IsParagraphStart)
```
tombogle: Now that setting prevBlock is pulled outside, these two if statements could be combined into 1. Since the body of the following if is also the same, they could all be combined into one if it didn't make it too unreadable.

phillip-hopper:
Review status: 1 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[Glyssen/BookScript.cs, line 113 \[r1\]](https://reviewable.io:443/reviews/sillsdev/glyssen/203#-KKKXX9XSCCyS_RRkwkL-r1-113:-KKOAa0c89ccO-ZP_0lN:-499146009) ([raw file](https://github.com/sillsdev/glyssen/blob/592b77c7017fae1e97475f26af90f0f3209f9653/Glyssen/BookScript.cs#L113)):*
<details><summary><i>Previously, tombogle (Tom Bogle) wrote…</i></summary>
> Now that setting prevBlock is pulled outside, these two if statements could be combined into 1. Since the body of the following if is also the same, they could all be combined into one if it didn't make it too unreadable.
</details>
Done.

---

*[Glyssen/BookScript.cs, line 121 \[r1\]](https://reviewable.io:443/reviews/sillsdev/glyssen/203#-KKKXX9XSCCyS_RRkwkL-r1-121:-KKOLWq93jc0nNVc0pwc:-499146009) ([raw file](https://github.com/sillsdev/glyssen/blob/592b77c7017fae1e97475f26af90f0f3209f9653/Glyssen/BookScript.cs#L121)):*
<details><summary><i>Previously, tombogle (Tom Bogle) wrote…</i></summary>
> Seems like it would be "better" to access the IsPoetic flag from the "TextProperties" of the style from the USFM stylesheet, rather than relying on the convention of starting these markers with the letter "q". Not sure how easily we can fluff up a stylsheet in this context (see USXParser for example).
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/glyssen/203)*

PR #223, opened 2016-08-01
PG-692: Options for extra-biblical parts

phillip-hopper:
PG-692: Options for extra-biblical parts

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/223)

### DIFF
--- a/Glyssen/Bundle/GlyssenDblTextMetadata.cs
+++ b/Glyssen/Bundle/GlyssenDblTextMetadata.cs
@@ -391,19 +391,26 @@ namespace Glyssen.Bundle
 		Required,
 	}
 
+	[Flags]
 	public enum ExtraBiblicalMaterialSpeakerOption
 	{
-		Narrator,
-		ActorOfEitherGender,
-		MaleActor,
-		FemaleActor,
-		Omitted,
+		Narrator = 1 << 0,
+		ActorOfEitherGender = 1 << 1,
+		MaleActor = 1 << 2,
+		FemaleActor = 1 << 3,
+		Omitted = 1 << 4,
+		MalePermitted = MaleActor | ActorOfEitherGender,
+		FemalePermitted = FemaleActor | ActorOfEitherGender,
+		NotNarratorOrOmitted = MaleActor | FemaleActor | ActorOfEitherGender
 	}
 
 	public class ProjectDramatizationPreferences
 	{
+		private ExtraBiblicalMaterialSpeakerOption test;
+
 		public ProjectDramatizationPreferences()
 		{
+			BookTitleAndChapterDramatization = ExtraBiblicalMaterialSpeakerOption.Narrator;
 			SectionHeadDramatization = ExtraBiblicalMaterialSpeakerOption.Omitted;
 			BookIntroductionsDramatization = ExtraBiblicalMaterialSpeakerOption.Omitted;
 		}
@@ -432,7 +439,7 @@ namespace Glyssen.Bundle
 
 		[XmlElement("BookIntroductionsDramatization")]
 		[DefaultValue(ExtraBiblicalMaterialSpeakerOption.Omitted)]
-		public ExtraBiblicalMaterialSpeakerOption BookIntroductionsDramatization { get; set; }
+		public ExtraBiblicalMaterialSpeakerOption BookIntroductionsDramatization{ get; set; }
 	}
 
 	[Flags]
--- a/Glyssen/Bundle/GlyssenDblTextMetadata.cs
+++ b/Glyssen/Bundle/GlyssenDblTextMetadata.cs
@@ -399,15 +399,11 @@ namespace Glyssen.Bundle
 		MaleActor = 1 << 2,
 		FemaleActor = 1 << 3,
 		Omitted = 1 << 4,
-		MalePermitted = MaleActor | ActorOfEitherGender,
-		FemalePermitted = FemaleActor | ActorOfEitherGender,
 		NotNarratorOrOmitted = MaleActor | FemaleActor | ActorOfEitherGender
 	}
 
 	public class ProjectDramatizationPreferences
 	{
-		private ExtraBiblicalMaterialSpeakerOption test;
-
 		public ProjectDramatizationPreferences()
 		{
 			BookTitleAndChapterDramatization = ExtraBiblicalMaterialSpeakerOption.Narrator;
@@ -439,7 +435,7 @@ namespace Glyssen.Bundle
 
 		[XmlElement("BookIntroductionsDramatization")]
 		[DefaultValue(ExtraBiblicalMaterialSpeakerOption.Omitted)]
-		public ExtraBiblicalMaterialSpeakerOption BookIntroductionsDramatization{ get; set; }
+		public ExtraBiblicalMaterialSpeakerOption BookIntroductionsDramatization { get; set; }
 	}
 
 	[Flags]
--- a/Glyssen/Character/CharacterVerseData.cs
+++ b/Glyssen/Character/CharacterVerseData.cs
@@ -52,12 +52,13 @@ namespace Glyssen.Character
 
 		public static StandardCharacter GetStandardCharacterType(string characterId)
 		{
-			if (String.IsNullOrEmpty(characterId))
+			if (string.IsNullOrEmpty(characterId))
 				return StandardCharacter.NonStandard;
 
-			int i = characterId.IndexOf("-", StringComparison.Ordinal);
+			var i = characterId.IndexOf("-", StringComparison.Ordinal);
 			if (i < 0)
 				return StandardCharacter.NonStandard;
+
 			switch (characterId.Substring(0, i + 1))
 			{
 				case kNarratorPrefix: return StandardCharacter.Narrator;
@@ -65,6 +66,7 @@ namespace Glyssen.Character
 				case kExtraBiblicalPrefix: return StandardCharacter.ExtraBiblical;
 				case kBookOrChapterPrefix: return StandardCharacter.BookOrChapter;
 			}
+
 			return StandardCharacter.NonStandard;
 		}
 
@@ -75,7 +77,21 @@ namespace Glyssen.Character
 
 		public static bool IsCharacterOfType(string characterId, StandardCharacter standardCharacterType)
 		{
-			return characterId.StartsWith(GetCharacterPrefix(standardCharacterType));
+			return characterId.StartsWith(GetCharacterPrefix(standardCharacterType), StringComparison.Ordinal);
+		}
+
+		public static bool IsCharacterExtraBiblical(string characterId)
+		{
+			if (IsCharacterOfType(characterId, StandardCharacter.BookOrChapter))
+				return true;
+
+			if (IsCharacterOfType(characterId, StandardCharacter.ExtraBiblical))
+				return true;
+
+			if (IsCharacterOfType(characterId, StandardCharacter.Intro))
+				return true;
+
+			return false;
 		}
 
 		public static string GetCharacterNameForUi(string characterId)
@@ -138,7 +154,7 @@ namespace Glyssen.Character
 			return characterId.Substring(characterId.Length - 3);
 		}
 
-		private static string GetCharacterPrefix(StandardCharacter standardCharacterType)
+		internal static string GetCharacterPrefix(StandardCharacter standardCharacterType)
 		{
 			switch (standardCharacterType)
 			{
--- a/Glyssen/Character/CharacterVerseData.cs
+++ b/Glyssen/Character/CharacterVerseData.cs
@@ -36,20 +36,6 @@ namespace Glyssen.Character
 			Intro,
 		}
 
-		public static bool IsCharacterStandard(string characterId, bool includeNarrator = true)
-		{
-			switch (GetStandardCharacterType(characterId))
-			{
-				case StandardCharacter.Narrator:
-					return includeNarrator;
-				case StandardCharacter.Intro:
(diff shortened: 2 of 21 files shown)


Glyssen/Bundle/GlyssenDblTextMetadata.cs
```diff
@@ -391,19 +391,26 @@ public enum PreferenceLevel
...
+		FemaleActor = 1 << 3,
+		Omitted = 1 << 4,
+		MalePermitted = MaleActor | ActorOfEitherGender,
+		FemalePermitted = FemaleActor | ActorOfEitherGender,
+		NotNarratorOrOmitted = MaleActor | FemaleActor | ActorOfEitherGender
 	}
 
 	public class ProjectDramatizationPreferences
 	{
+		private ExtraBiblicalMaterialSpeakerOption test;
```
tombogle: field test is not used.

Glyssen/Bundle/GlyssenDblTextMetadata.cs
```diff
@@ -391,19 +391,26 @@ public enum PreferenceLevel
...
-		MaleActor,
-		FemaleActor,
-		Omitted,
+		Narrator = 1 << 0,
+		ActorOfEitherGender = 1 << 1,
+		MaleActor = 1 << 2,
+		FemaleActor = 1 << 3,
+		Omitted = 1 << 4,
+		MalePermitted = MaleActor | ActorOfEitherGender,
+		FemalePermitted = FemaleActor | ActorOfEitherGender,
```
tombogle: MaklePermitted and FemalePermitted are not used. Are they there for some future (planned) use?

Glyssen/Bundle/GlyssenDblTextMetadata.cs
```diff
@@ -432,7 +439,7 @@ public ProjectDramatizationPreferences()
 
 		[XmlElement("BookIntroductionsDramatization")]
 		[DefaultValue(ExtraBiblicalMaterialSpeakerOption.Omitted)]
-		public ExtraBiblicalMaterialSpeakerOption BookIntroductionsDramatization { get; set; }
+		public ExtraBiblicalMaterialSpeakerOption BookIntroductionsDramatization{ get; set; }
```
tombogle: This change (removal of space) appears to have been done inadvertently.

Glyssen/Character/CharacterVerseData.cs
```diff
@@ -75,7 +77,21 @@ public static string GetStandardCharacterId(string bookId, StandardCharacter sta
...
+			if (IsCharacterOfType(characterId, StandardCharacter.BookOrChapter))
+				return true;
+
+			if (IsCharacterOfType(characterId, StandardCharacter.ExtraBiblical))
+				return true;
+
+			if (IsCharacterOfType(characterId, StandardCharacter.Intro))
+				return true;
+
+			return false;
```
tombogle: Based on my speed testing, this method (which is not used) is much faster than our current implementation of IsCharacterStandard. I proposed we implement two methods:
		public static bool IsCharacterStandard(string characterId)
		{
			return IsCharacterOfType(characterId, StandardCharacter.Narrator) ||
				IsCharacterOfType(characterId, StandardCharacter.BookOrChapter) ||
				IsCharacterOfType(characterId, StandardCharacter.ExtraBiblical) ||
				IsCharacterOfType(characterId, StandardCharacter.Intro);
				// We could call IsCharacterExtraBiblical instead of the last three lines of this if,
				// but this is speed-critical code and the overhead of the extra method call is
				// expensive.
		}

		public static bool IsCharacterExtraBiblical(string characterId)
		{
			return IsCharacterOfType(characterId, StandardCharacter.BookOrChapter) ||
				IsCharacterOfType(characterId, StandardCharacter.ExtraBiblical) ||
				IsCharacterOfType(characterId, StandardCharacter.Intro);
		}
IsCharacterExtraBiblical should be called any place in the code where we currently call IsCharacterStandard with a false parameter. Any of the existing single-parameter calls (true) will continue to compile just fine using the single parameter version of that method.

Glyssen/Rules/CharacterGroupGenerator.cs
```diff
@@ -914,13 +1021,118 @@ public static void DistributeBooksAmongNarratorGroups(List<AuthorStats> authorSt
 				}
 			}
 
-			private void AssignExtraBiblicalCharactersToExtraBiblicalGroups(List<CharacterDetail> includedCharacterDetails)
+			private void AssignExtraBiblicalCharactersToExtraBiblicalGroups(List<CharacterDetail> includedCharacterDetails,
+				ProjectDramatizationPreferences dramatizationPreferences)
+			{
+				var narratorPrefix = CharacterVerseData.GetCharacterPrefix(CharacterVerseData.StandardCharacter.Narrator);
```
tombogle: Rather than passing narrator prefix down through all these layers of private methods (including possible paths where it might never be needed), why not just get it once at the one line of code that uses it?

Glyssen/Rules/CharacterGroupGenerator.cs
```diff
@@ -914,13 +1021,118 @@ public static void DistributeBooksAmongNarratorGroups(List<AuthorStats> authorSt
...
+
+				// get standard characters not yet assigned
+				foreach (var character in includedCharacterDetails.Where(c => !Groups.Any(g => g.CharacterIds.Contains(c.CharacterId))).ToList())
+				{
+					var characterType = character.StandardCharacterType;
+					switch (characterType)
+					{
+						case CharacterVerseData.StandardCharacter.BookOrChapter:
```
tombogle: I think I would prefer to have each case statement merely set a local variable for the group, followed by a single call to AddExtraBiblicalToGroupOrNarratorGroup using that variable as a parameter, rather than repeating an otherwise-identical method call three times. It would seem to make the the code more compact and reduce the maintenance load. (This would require a default: case with an early return.)

Glyssen/Rules/CharacterGroupGenerator.cs
```diff
@@ -914,13 +1021,118 @@ public static void DistributeBooksAmongNarratorGroups(List<AuthorStats> authorSt
...
+							break;
+					}
+				}
+			}
+
+			private void AddExtraBiblicalToGroupOrNarratorGroup(string characterId, CharacterGroup group,
+				CharacterVerseData.StandardCharacter characterType, string narratorPrefix, ProjectDramatizationPreferences dramatizationPreferences)
 			{
-				if (ExtraBiblicalGroup == null)
+				if (group != null)
```
tombogle: Might want to just put this up in the previous method (after taking my other suggestion there) to avoid the cost of a method call and pushing a bunch of parameters that are not needed in this case.

Glyssen/Bundle/GlyssenDblTextMetadata.cs
phillip-hopper: Removed.

Glyssen/Bundle/GlyssenDblTextMetadata.cs
phillip-hopper: Removed.

Glyssen/Bundle/GlyssenDblTextMetadata.cs
phillip-hopper: Fixed.

Glyssen/Rules/CharacterGroupGenerator.cs
```diff
@@ -914,13 +1021,118 @@ public static void DistributeBooksAmongNarratorGroups(List<AuthorStats> authorSt
...
+
+					case CharacterVerseData.StandardCharacter.Intro:
+						dramatizationOption = dramatizationPreferences.BookIntroductionsDramatization;
+						break;
+				}
+
+				if (dramatizationOption == ExtraBiblicalMaterialSpeakerOption.Omitted)
+					return null;
+
+				var narratorId = narratorPrefix + characterId.Split(new[] { '-' }, 2)[1];
```
tombogle: I wonder if there is a clearer/faster way to write this that wouldn't involve allocating an array, creating two new strings, array element access, and then creating another new string. Since for each type of character (as determined in the above switch), we can know the exact character offset where we expect to find the dash, it *might* be worth caching and using that information to create the new string for the narrator ID. ***OR*** since the thing following the dash is always the same length, we could just use that to pick off the book ID portion.

Glyssen/Rules/CharacterGroupGenerator.cs
```diff
@@ -914,13 +1021,118 @@ public static void DistributeBooksAmongNarratorGroups(List<AuthorStats> authorSt
...
+				List<CharacterGroup> possibleGroups;
+
+				if (dramatizationOption == ExtraBiblicalMaterialSpeakerOption.Narrator)
+					possibleGroups = NarratorGroups.Where(g => g.CharacterIds.Contains(narratorId)).ToList();
+				else
+					possibleGroups = NarratorGroups.Where(g => !g.CharacterIds.Contains(narratorId)).ToList();
+
+				if (possibleGroups.Count == 0)
+					possibleGroups = NarratorGroups.Except(possibleGroups).ToList();
```
tombogle: No need to use Except since possibleGroups has no items.

Glyssen/Rules/CharacterGroupGenerator.cs
```diff
@@ -914,13 +1021,118 @@ public static void DistributeBooksAmongNarratorGroups(List<AuthorStats> authorSt
...
+
+				CharacterGroup narratorGroup = null;
+
+				// this should never happen, so throw an exception if it does
+				if (possibleGroups.Count == 0)
```
tombogle: This could be moved inside body of above if (possibleGroup.Count == 0) to save unnecessary re-check.

Glyssen/Rules/CharacterGroupGenerator.cs
```diff
@@ -914,13 +1021,118 @@ public static void DistributeBooksAmongNarratorGroups(List<AuthorStats> authorSt
...
+					throw new NullReferenceException(string.Format("No group was found for character '{0}'", characterId));
+
+				if (possibleGroups.Count == 1)
+				{
+					return possibleGroups[0];
+				}
+
+				// get the narrator group with the least amount of speaking
+				var leastEstimatedHours = double.MaxValue;
```
tombogle: Use this Linq expression:
narratorGroup = possibleGroups.MinBy(g => g.EstimatedHours);

Glyssen/Rules/CharacterGroupGenerator.cs
```diff
@@ -914,13 +1021,118 @@ public static void DistributeBooksAmongNarratorGroups(List<AuthorStats> authorSt
...
+						narratorGroup = cg;
+
+						// is this a good enough match?
+						if (leastEstimatedHours < 0.01)
+							break;
+					}
+				}
+
+				// this should never happen, so throw an exception if it does
+				if (narratorGroup == null)
```
tombogle: With about Linq expression, it should be obvious that this can't happen, so you can remove this check.

Glyssen/Rules/Proximity.cs
```diff
@@ -39,28 +70,29 @@ public MinimumProximity CalculateMinimumProximity(ISet<string> characterIdsToCal
...
-					standardCharacterIdsForBook.Clear();
-					standardCharacterIdsForBook.Add(CharacterVerseData.GetStandardCharacterId(book.BookId, CharacterVerseData.StandardCharacter.BookOrChapter));
-					standardCharacterIdsForBook.Add(CharacterVerseData.GetStandardCharacterId(book.BookId, CharacterVerseData.StandardCharacter.ExtraBiblical));
-					standardCharacterIdsForBook.Add(CharacterVerseData.GetStandardCharacterId(book.BookId, CharacterVerseData.StandardCharacter.Intro));
+				var treatAsSameCharacter = m_considerSameExtrabiblicalCharacter[book];
 
-					if (standardCharacterIdsForBook.Any(characterIdsToCalculate.Contains))
-						standardCharacterIdsToTreatAsOne = new HashSet<string>();
-				}
+				////if (!treatStandardNonScriptureCharactersAsDistinct)
```
tombogle: Unless you think this commented-out code is useful for future reference, let's delete it.

Glyssen/Character/CharacterVerseData.cs
phillip-hopper: Implemented.

Glyssen/Rules/CharacterGroupGenerator.cs
phillip-hopper: Implemented.

Glyssen/Rules/CharacterGroupGenerator.cs
phillip-hopper: Implemented.

Glyssen/Rules/CharacterGroupGenerator.cs
phillip-hopper: Implemented.

Glyssen/Rules/CharacterGroupGenerator.cs
phillip-hopper: Implemented.

Glyssen/Rules/CharacterGroupGenerator.cs
phillip-hopper: Fixed.

Glyssen/Rules/CharacterGroupGenerator.cs
phillip-hopper: Implemented.

tombogle:
Superseded by PR 231

PR #226, opened 2016-08-09
PG-788: Font applied in settings should be applied in the export file

phillip-hopper:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/226)

### DIFF
--- a/Glyssen/ProjectExporter.cs
+++ b/Glyssen/ProjectExporter.cs
@@ -338,7 +338,8 @@ namespace Glyssen
 			using (var xls = new ExcelPackage(new FileInfo(path)))
 			{
 				var sheet = xls.Workbook.Worksheets.Add("Script");
-				sheet.Cells["A1"].LoadFromArrays(dataArray);
+				var firstCell = sheet.Cells["A1"];
+				firstCell.LoadFromArrays(dataArray);
 
 				ColorizeAnnotations(sheet);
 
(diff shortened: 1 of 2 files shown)


tombogle:
I don't think we should apply the vernacular font to the localized character ID. (Ultimately, we might well need to have a place/way that the font family and size can be specified for each localized language, but we can't safely assume it should be the same as the vernacular.)
We *do* need to use the font info from the reference text(s) for those columns.

Glyssen/ProjectExporter.cs
```diff
@@ -357,18 +358,34 @@ private void GenerateExcelFile(string path, IEnumerable<List<object>> data)
 				sheet.Column(columnNum++).AutoFit(2d, sheet.DefaultColWidth); // verse
 				sheet.Column(columnNum++).AutoFit(2d, 20d); // character ID
 
-				// add a column for the localized character id
```
tombogle: I don't think we want to remove this code. If we really don't want to auto-fit this column (and why not?), it doesn't feel like it has anything to do with this Jira issue. In any case, I think we would need to increment columnNum.

phillip-hopper: Undone.

Glyssen/ReferenceText.cs
```diff
@@ -533,5 +533,15 @@ protected override string ProjectFolder
 				return m_projectFolder;
 			}
 		}
+
```
tombogle: Note: This has already been added (elsewhere in the file) in master, so this will result in a build error when merged.

PR #227, opened 2016-08-10
PG-769: Generate folder of labelled blank clip files

phillip-hopper:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/227)


Glyssen/ProjectExporter.cs
```diff
@@ -310,6 +337,53 @@ internal DataTable GeneratePreviewTable()
 			}
 		}
 
+		private void GenerateClipFiles(string directoryPath)
+		{
+			const int kLineNumIndex = 0;
+			const int kBookIndex = 2;
```
tombogle: We should use the ExportColumn enum instead of these constants. These will be off by one if the Actor column is present. I'm checking with Duane, but I think the option of generating clip files should only be available if there are actor assignments.

phillip-hopper: Implemented.

tombogle:
I have added two subtasks to this Jira issue. I think we can wait on the one to add metadata, but we should try to do the other one (adding the column to the output with a clickable link to the file). Also, Duane confirmed that this option should only be available in the export dialog if the user has assigned at least one voice actor.

PR #258, opened 2016-10-20
Update publish path

irahopkinson:
The organization of the directory used by software.sil.org has changed to have an extra "r" directory.

Adding sudo -u vu2004 to the --rsync-path will cause the uploaded files to be owned by vu2004 which is the same user that would operate FTP. This allows product teams to use a tool like Filezilla to delete old files from time to time.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/258)


tombogle:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/glyssen/258)*

PR #261, opened 2016-11-03
Make ILRepack work

ermshiperete:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/261)


tombogle:
We've decided to go with the same approach as in HearThis. Hopefully more straightforward than ILRepack.

PR #322, opened 2017-01-25
PG-894 Change AssignCharacterDlg in Rainbow mode.  If a default character is…

tombogle:
… set for a Section Head, it cannot be edited so the drop down list of characters disappears for that line in the reference text pane.  Owner drawn.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/322)


tombogle:
I don't think we want the Designer file changes.

PR #392, opened 2017-06-14
Refactor to create GlyssenShared and create GlyssenScript class

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/392)


JohnThomson:
I think I was asked to review this, so I've done my best as someone not familiar with the code base or even with some of the assumed terminology. Feel free to ignore anything unhelpful.

---

Reviewed 33 of 80 files at r1, 66 of 68 files at r2.
Review status: all files reviewed at latest revision, 25 unresolved discussions.

---

*[Glyssen/Block.cs, line 22 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonHnKVc2LTpO78Hvbo:-KonHnKWY3mmoXkTosWx:bdzuh3j) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/Block.cs#L22)):*
> ```Smalltalk
> {
> 	[XmlRoot("block")]
> 	public class Block
> ```

Especially being promoted to a shared library, a class needs a comment.

---

*[Glyssen/Block.cs, line 250 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonI31mO49SNExSi6Vc:-KonI31mO49SNExSi6Vd:b1mmwws) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/Block.cs#L250)):*
> ```Smalltalk
> 		}
> 
> 		public string GetPrimaryReferenceText(bool scriptureTextOnly = false)
> ```

Changing from a public property to a method will (I think) affect XML serialization, if that is what the XMlRoot annotation is about. Is that intentional?
It may be that "Primary Reference Text" is a sufficiently prominent Glyssen concept that it is described once in some overview doc. Otherwise, I'd want a comment explaining it., particularly how it is modified by the scriptureTextOnly flag.

---

*[Glyssen/BookBlockIndices.cs, line 8 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonJY6V_u01075Oyxz6:-KonJY6V_u01075Oyxz7:b-yqa0v1) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/BookBlockIndices.cs#L8)):*
> ```Smalltalk
> {
> 	[XmlRoot]
> 	public class BookBlockIndices : IEquatable<BookBlockIndices>, IComparable<BookBlockIndices>
> ```

There's no good way to compare this with the old, deleted class. If something other than moving it changed, maybe you could point it out with a review comment?
As before, a public class, especially in a shared library, should have a  helpful comment. (I won't bother to go on saying this but please check it for all classes in the shared library.)

---

*[Glyssen/DataMigrator.cs, line 141 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonKCVXxG3GSczC8LJB:-KonKCVXxG3GSczC8LJC:bxu67la) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/DataMigrator.cs#L141)):*
> ```Smalltalk
> 					foreach (var pgProjFile in Project.AllRecordingProjectFolders.SelectMany(d => Directory.GetFiles(d, "*" + kOldProjectExtension)))
> 					{
> 						var newName = Path.ChangeExtension(pgProjFile, Constants.kProjectFileExtension);
> ```

I suppose this is necessary to make the GlyssenShared component self-contained, but otherwise, Project feels like a good home for the constant extension for projects.

---

*[Glyssen/Project.cs, line 52 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonKhKFA_feb0DcFmqG:-KonKhKFA_feb0DcFmqH:b-tq7sks) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/Project.cs#L52)):*
> ```Smalltalk
> 		private const double kQuotePercent = 0.65;
> 
> 		// Ideally these will be settings
> ```

I suppose key strokes per hour could even be measured. But I doubt it would be constant :-)
But if you expect it to become a setting why declare it as a constant?

---

*[Glyssen/Project.cs, line 298 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonMYShJTYhwT87ufEy:-KonMYShJTYhwT87ufEz:b-hs3j5k) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/Project.cs#L298)):*
> ```Smalltalk
> 		}
> 
> 		public bool IncludeCharacter(string characterId)
> ```

Feels to me as though this method might be more at home on  DramatizationPreferences (could be delegated from here if necessary).

---

*[Glyssen/Project.cs, line 1742 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonNa7TlPm0w-fmCES9:-KonNa7TlPm0w-fmCESA:b-8co9ce) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/Project.cs#L1742)):*
> ```Smalltalk
> 		}
> 
> 		public double GetEstimatedRecordingTime()
> ```

This is obviously not a change, but it seems a strange way to estimate. Why would the time to record something be better correlated with keystrokes than simply the number of code points? Or possibly better yet the number of non-diacritic  word-forming code points?

---

*[Glyssen/ProjectExporter.cs, line 72 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonOagQMY5-8CW2By_8:-KonOagQMY5-8CW2By_9:b-9i0kbh) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/ProjectExporter.cs#L72)):*
> ```Smalltalk
> 
> 		/// <summary>
> 		/// True to exclude verse numbers and annotations; default is false
> ```

Including book introductions? (Is that an 'annotation'? It's not Scripture. What about 'scriptural' annotations like 'selah' or the introductions at the start of many Psalms?)

---

*[Glyssen/Bundle/GlyssenDblTextMetadata.cs, line 90 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonQ5w05NqOyjshU5bD:-KonQ5w05NqOyjshU5bE:bbvc9b6) ([raw file](https://github.com/sillsdev/glyssen/blob/15a1c0029a0c3842d8cf9dbf7f2ddfdf621f3732/Glyssen/Bundle/GlyssenDblTextMetadata.cs#L90)):*
> ```Smalltalk
> 		/// </summary>
> 		[XmlAttribute("modifieddate")]
> 		public DateTime LastModified { get; set; }
> ```

What's the implication for XML serialization of removing a previous property? Maybe it's OK because Glyssen hasn't been released?

---

*[Glyssen/Dialogs/AssignCharacterViewModel.cs, line 9 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonQsoNVx7P29zUUd1y:-KonQsoNVx7P29zUUd1z:b-24skmn) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/Dialogs/AssignCharacterViewModel.cs#L9)):*
> ```Smalltalk
> using Glyssen.Character;
> using Glyssen.Shared;
> using Glyssen.Shared.Bundle;
> ```

This is as good a place as any to say that I'm puzzled by the number of files that needs new using declarations though no old one was removed and there are no other changes. Of course it's possible that some unchanged code uses something that used to be in the other dependencies but is now in the new component...

---

*[GlyssenShared/IReferenceTextIdentifier.cs, line 5 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonS2FKJXzV7qMp6mcP:-KonS2FKJXzV7qMp6mcQ:b-ncyt7s) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/IReferenceTextIdentifier.cs#L5)):*
> ```Smalltalk
> namespace Glyssen.Shared
> {
> 	public interface IReferenceTextIdentifier
> ```

Not keen on this class name, based on the property names. How can an identifier be Missing? It feels wrong for an identifier to know where to find the thing identified (ProjectFolder). The other two, also, feel more like properties of the reference text than of its identifier.

---

*[GlyssenShared/Bundle/GlyssenDblTextMetadataBase.cs, line 7 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonT_tOmhy3GQMo5a6C:-KonT_tOmhy3GQMo5a6D:b-l7x9cu) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Bundle/GlyssenDblTextMetadataBase.cs#L7)):*
> ```Smalltalk
> namespace Glyssen.Shared.Bundle
> {
> 	public interface IReadOnlyGlyssenDblTextMetadata
> ```

When you comment this and similar classes, it would be helpful to note that Dbl is not short for 'double'. (I think I finally figured out it means Digital Bible Library?)

---

*[GlyssenShared/Script/GlyssenScript.cs, line 9 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonVBWJghq42UBtO-F5:-KonVBWJghq42UBtO-F6:buya4rk) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L9)):*
> ```Smalltalk
> {
> 	[XmlRoot("glyssenscript")]
> 	public class GlyssenScript : GlyssenDblTextMetadataBase
> ```

It feels odd that the text itself is a kind of metadata. The text ought to be the data itself and HAVE metadata. Of course as an object it likely has the metadata attributes, and the relationship may be required for serialization to work as desired. Some discussion of this in class comment might be worthwhile.

---

*[GlyssenShared/Script/GlyssenScript.cs, line 40 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonVkMm3giwAsd95HT2:-KonVkMm3giwAsd95HT3:b5izy68) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L40)):*
> ```Smalltalk
> 	}
> 
> 	public class Script
> ```

You seem to have been working on moving classes into their own files...these ones are admittedly very trivial...

---

*[GlyssenShared/Script/GlyssenScript.cs, line 42 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonVurLxbLzFTZARBcz:-KonVurLxbLzFTZARBd-:b-vckhkm) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L42)):*
> ```Smalltalk
> 	public class Script
> 	{
> 		[XmlElement(ElementName = "book")]
> ```

Is this the name for each item in the list? It would seem odd to use the singular for the list as a whole.

---

*[GlyssenShared/Script/GlyssenScript.cs, line 49 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonWBBDprDbSXgd2lQP:-KonWBBDprDbSXgd2lQQ:be7bsxd) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L49)):*
> ```Smalltalk
> 	{
> 		[XmlAttribute("id")]
> 		public string Id { get; set; }
> ```

Wants doc...1? gen? GEN? Genesis?

---

*[GlyssenShared/Script/GlyssenScript.cs, line 57 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonWMSRPua2sztkQPhF:-KonWMSRPua2sztkQPhG:bruzd6u) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L57)):*
> ```Smalltalk
> 	public class ScriptChapter
> 	{
> 		[XmlAttribute("id")]
> ```

Wants doc...are you using zero for introduction, like HearThis?

---

*[GlyssenShared/Script/GlyssenScript.cs, line 67 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonWXMyZGtem2_tm0wq:-KonWXMyZGtem2_tm0wr:b-70k4ng) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L67)):*
> ```Smalltalk
> 	{
> 		[XmlAttribute("id")]
> 		public int Id { get; set; }
> ```

Wants doc...are blocks in chapter numbered from 0, or 1? Can there be gaps?

---

*[GlyssenShared/Script/GlyssenScript.cs, line 77 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonWixNTZwfY0nBUb5T:-KonWixNTZwfY0nBUb5U:b-v8henj) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L77)):*
> ```Smalltalk
> 		public string Character { get; set; }
> 		[XmlAttribute("delivery")]
> 		public string Delivery { get; set; }
> ```

Meaning is not obvious...Delivery as in something to do with how an actor delivers lines?

---

*[GlyssenShared/Script/GlyssenScript.cs, line 79 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonWuxq52CwCO1fuMi4:-KonWuxq52CwCO1fuMi5:beyqcv9) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L79)):*
> ```Smalltalk
> 		public string Delivery { get; set; }
> 		[XmlAttribute("file")]
> 		public string File { get; set; }
> ```

Where the audio is stored? How about AudioFile? Also doc what it is and how complete...1.wav? Genesis/1/1.wav? C:/Users/Bogle/.../MyProject/Genesis/1/1.wav?

---

*[GlyssenShared/Script/GlyssenScript.cs, line 82 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonXTgTPOl_9MHK4x36:-KonXTgTPOl_9MHK4x37:b58mdae) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L82)):*
> ```Smalltalk
> 
> 		[XmlElement("vern")]
> 		public Vernacular Vernacular { get; set; }
> ```

Wish we could make this more prominent somehow...this is the actual text to be recorded, right? The property everything else is about. Simply 'Text' is a pretty standard name for the main text content of an object that has such a primary content attribute. (Don't feel any pressure to change this...just wondering.)

---

*[GlyssenTests/Dialogs/CastSizePlanningViewModelTests.cs, line 446 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonZDkUHERCkfO0F7sL:-KonZDkUHERCkfO0F7sM:b-5ffhpq) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenTests/Dialogs/CastSizePlanningViewModelTests.cs#L446)):*
> ```Smalltalk
> 
> 	[TestFixture]
> 	class CalculateMinimumCastSizesForOldTestamentBasedOnKunaSanBlas
> ```

How can you have a TestFixture that isn't public? Do these tests even get run? I think the class wants its own file.

---

*[Reference Text Utility/RefTextUtilityForm.cs, line 154 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonZbzR3nEE4Sta3PKe:-KonZbzR3nEE4Sta3PKf:b-l5blu4) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Reference Text Utility/RefTextUtilityForm.cs#L154)):*
> ```Smalltalk
> 			{
> 				Data.FilterBy(languagesToDiff);
> 				RefTextDevUtilities.ReferenceTextUtility.ProcessReferenceTextData(mode, Data, dummy:null);
> ```

I don't recall seeing this declared, but what's an argument doing called 'dummy'?

---

*[RefTextDevUtilities/ReferenceTextUtility.cs, line 212 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-Kon_3_IhKtgZXv2jdgY:-Kon_3_IhKtgZXv2jdgZ:b62x20o) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/RefTextDevUtilities/ReferenceTextUtility.cs#L212)):*
> ```Smalltalk
> 		/// <param name="data"></param>
> 		/// <param name="dummy"></param>
> 		public static void ProcessReferenceTextData(Mode mode, ReferenceTextData data, string dummy)
> ```

Very confusing. For one thing it's NOT an override, so why say it is and have an extra argument deliberately not used? Needs at least a better explanation.

---

*[RefTextDevUtilities/ReferenceTextUtility.cs, line 214 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-Kon_KWC6GjZRZieu-Si:-Kon_KWC6GjZRZieu-Sj:b-3nuo3x) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/RefTextDevUtilities/ReferenceTextUtility.cs#L214)):*
> ```Smalltalk
> 		public static void ProcessReferenceTextData(Mode mode, ReferenceTextData data, string dummy)
> 		{
> 			ProcessReferenceTextData(mode, data, (Func<ReferenceText>)null);
> ```

You can now just omit the unwanted optional param

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-:-Kon__Zz7EL1fheZY4S-:b-w04n4b)*

andrew-polk:
Thanks. I know that was a bear to review (and re-review)...

---

Review status: all files reviewed at latest revision, 25 unresolved discussions.

---

*[Glyssen/Block.cs, line 22 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonHnKVc2LTpO78Hvbo:-Koqpe7rtM1Bp-8dfjrP:b-jvu2ww) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/Block.cs#L22)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Especially being promoted to a shared library, a class needs a comment.
</blockquote></details>

This particular class is not being promoted. Glyssen historically hasn't bothered with class comments (not saying we shouldn't have). But I'll compromise by adding comments for classes in GlyssenShared.

---

*[Glyssen/Block.cs, line 250 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonI31mO49SNExSi6Vc:-KoqppCgQeEo6WevaxiX:bbn7x94) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/Block.cs#L250)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Changing from a public property to a method will (I think) affect XML serialization, if that is what the XMlRoot annotation is about. Is that intentional?
It may be that "Primary Reference Text" is a sufficiently prominent Glyssen concept that it is described once in some overview doc. Otherwise, I'd want a comment explaining it., particularly how it is modified by the scriptureTextOnly flag.
</blockquote></details>

Because the property had no setter, it was not previously serialized. I did add a test proving that the thing we actually do serialize (ReferenceBlocks) is getting round-tripped properly.

---

*[Glyssen/BookBlockIndices.cs, line 8 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonJY6V_u01075Oyxz6:-KosIuU-kY4NqMatkAg2:b-oyb61q) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/BookBlockIndices.cs#L8)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

There's no good way to compare this with the old, deleted class. If something other than moving it changed, maybe you could point it out with a review comment?
As before, a public class, especially in a shared library, should have a  helpful comment. (I won't bother to go on saying this but please check it for all classes in the shared library.)
</blockquote></details>

No change. Thanks for checking.

---

*[Glyssen/DataMigrator.cs, line 141 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonKCVXxG3GSczC8LJB:-KoqqOaR_rNb_8iDugGC:bwpp1c4) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/DataMigrator.cs#L141)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

I suppose this is necessary to make the GlyssenShared component self-contained, but otherwise, Project feels like a good home for the constant extension for projects.
</blockquote></details>

Yes, necessary to make GlyssenShared self-contained

---

*[Glyssen/Project.cs, line 52 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonKhKFA_feb0DcFmqG:-Koqqh3bnmDu0TH6GMXR:bv6megq) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/Project.cs#L52)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

I suppose key strokes per hour could even be measured. But I doubt it would be constant :-)
But if you expect it to become a setting why declare it as a constant?
</blockquote></details>

Because that is the best we have for now. : )
With this change, I'm just moving it and adding the comment explaining where we want to go eventually. I'll change the comment to "enhance" to be more clear.

---

*[Glyssen/Project.cs, line 298 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonMYShJTYhwT87ufEy:-Koqvbr3TbuzIxYo60nP:b-s5g0dm) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/Project.cs#L298)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Feels to me as though this method might be more at home on  DramatizationPreferences (could be delegated from here if necessary).
</blockquote></details>

Thanks. I think this must have happened at some intermediate stage in the refactoring which was subsequently changed. I've moved it back to ProjectDramatizationPreferences

---

*[Glyssen/Project.cs, line 1742 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonNa7TlPm0w-fmCES9:-Koqvtv007VzLAHqhILn:b-38kptg) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/Project.cs#L1742)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

This is obviously not a change, but it seems a strange way to estimate. Why would the time to record something be better correlated with keystrokes than simply the number of code points? Or possibly better yet the number of non-diacritic  word-forming code points?
</blockquote></details>

key strokes = code points
Key strokes was a term we used to try to distinguish from "character" which has multiple meanings in this context. Perhaps letters (not quite accurate either) or code points would have been better.

---

*[Glyssen/ProjectExporter.cs, line 72 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonOagQMY5-8CW2By_8:-KoqwNRNFqXhCK3nUpzm:bz37zhw) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/ProjectExporter.cs#L72)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Including book introductions? (Is that an 'annotation'? It's not Scripture. What about 'scriptural' annotations like 'selah' or the introductions at the start of many Psalms?)
</blockquote></details>

Annotation means something very specific here. But I see your point that "Scripture" only is not accurate. I'll try to make it more clear.

---

*[Glyssen/Bundle/GlyssenDblTextMetadata.cs, line 90 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonQ5w05NqOyjshU5bD:-KoquIMZojFbvbGrZz5n:b-98mhfq) ([raw file](https://github.com/sillsdev/glyssen/blob/15a1c0029a0c3842d8cf9dbf7f2ddfdf621f3732/Glyssen/Bundle/GlyssenDblTextMetadata.cs#L90)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

What's the implication for XML serialization of removing a previous property? Maybe it's OK because Glyssen hasn't been released?
</blockquote></details>

They were moved to a base class.

---

*[Glyssen/Dialogs/AssignCharacterViewModel.cs, line 9 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonQsoNVx7P29zUUd1y:-KoqzC_K29i4S3ue0GL0:bu0lov1) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/Dialogs/AssignCharacterViewModel.cs#L9)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

This is as good a place as any to say that I'm puzzled by the number of files that needs new using declarations though no old one was removed and there are no other changes. Of course it's possible that some unchanged code uses something that used to be in the other dependencies but is now in the new component...
</blockquote></details>

Yes, that is the case. Many things now reference classes in GlyssenShared which previously just referenced classes in Glyssen.
But your comment prompted me to look, and there were a bunch of Glyssen.Shared.Bundle using declarations I was able to get rid of. Thanks.

---

*[GlyssenShared/IReferenceTextIdentifier.cs, line 5 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonS2FKJXzV7qMp6mcP:-KorxIQn_JW9W4dxMtff:b-r91dx8) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/IReferenceTextIdentifier.cs#L5)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Not keen on this class name, based on the property names. How can an identifier be Missing? It feels wrong for an identifier to know where to find the thing identified (ProjectFolder). The other two, also, feel more like properties of the reference text than of its identifier.
</blockquote></details>

Good point. This was the result of extracting the bits of an unfortunately-named class to an interface so I could use it without tying in other dependencies...
I'll try to come up with a better name, probably for the class, too.

---

*[GlyssenShared/Bundle/GlyssenDblTextMetadataBase.cs, line 7 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonT_tOmhy3GQMo5a6C:-KosL7OMMIFoF6uADdkg:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Bundle/GlyssenDblTextMetadataBase.cs#L7)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

When you comment this and similar classes, it would be helpful to note that Dbl is not short for 'double'. (I think I finally figured out it means Digital Bible Library?)
</blockquote></details>

Done.

---

*[GlyssenShared/Script/GlyssenScript.cs, line 9 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonVBWJghq42UBtO-F5:-KosEgTINWaRaeDGav1I:bq2ho8j) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L9)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

It feels odd that the text itself is a kind of metadata. The text ought to be the data itself and HAVE metadata. Of course as an object it likely has the metadata attributes, and the relationship may be required for serialization to work as desired. Some discussion of this in class comment might be worthwhile.
</blockquote></details>

I agree it seems odd... but yes, it is done for serialization reasons.
I've added a comment.

---

*[GlyssenShared/Script/GlyssenScript.cs, line 40 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonVkMm3giwAsd95HT2:-Kos3GifTiJAaOE0Mxr2:bhy40e5) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L40)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

You seem to have been working on moving classes into their own files...these ones are admittedly very trivial...
</blockquote></details>

Yeah, I think I'll leave these together.

---

*[GlyssenShared/Script/GlyssenScript.cs, line 42 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonVurLxbLzFTZARBcz:-KorzbQ60ucjB6DJrIdS:b-ynq81g) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L42)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Is this the name for each item in the list? It would seem odd to use the singular for the list as a whole.
</blockquote></details>

No, this construct makes each item in the list a "book" without needing adding an unnecessary level.

---

*[GlyssenShared/Script/GlyssenScript.cs, line 49 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonWBBDprDbSXgd2lQP:-Kos8tr1VT70nglE_OOD:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L49)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Wants doc...1? gen? GEN? Genesis?
</blockquote></details>

Done.

---

*[GlyssenShared/Script/GlyssenScript.cs, line 57 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonWMSRPua2sztkQPhF:-Kos80wBEq5e7REB0BPw:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L57)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Wants doc...are you using zero for introduction, like HearThis?
</blockquote></details>

Done.

---

*[GlyssenShared/Script/GlyssenScript.cs, line 67 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonWXMyZGtem2_tm0wq:-KosGArTQ2lEoV3LXC48:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L67)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Wants doc...are blocks in chapter numbered from 0, or 1? Can there be gaps?
</blockquote></details>

Done.

---

*[GlyssenShared/Script/GlyssenScript.cs, line 77 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonWixNTZwfY0nBUb5T:-Kos8wmgDfLhZ5WHUkLr:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L77)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Meaning is not obvious...Delivery as in something to do with how an actor delivers lines?
</blockquote></details>

Done.

---

*[GlyssenShared/Script/GlyssenScript.cs, line 79 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonWuxq52CwCO1fuMi4:-Kos8xhUh8_pGJFu0pop:bpo8kne) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L79)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Where the audio is stored? How about AudioFile? Also doc what it is and how complete...1.wav? Genesis/1/1.wav? C:/Users/Bogle/.../MyProject/Genesis/1/1.wav?
</blockquote></details>

I think I'm going to just remove this. I think Tom put it in the data model based on the assumption HearThis would record into the files Glyssen created. So I see no reason for it now.

---

*[GlyssenShared/Script/GlyssenScript.cs, line 82 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonXTgTPOl_9MHK4x36:-Kos4iscWg6JjD1kDptl:bdy6377) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L82)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Wish we could make this more prominent somehow...this is the actual text to be recorded, right? The property everything else is about. Simply 'Text' is a pretty standard name for the main text content of an object that has such a primary content attribute. (Don't feel any pressure to change this...just wondering.)
</blockquote></details>

Are you suggesting we may want to change the xml? Or just internally?

---

*[GlyssenTests/Dialogs/CastSizePlanningViewModelTests.cs, line 446 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonZDkUHERCkfO0F7sL:-Kos08IIMxCkAfQhdzh6:b-pyfko9) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenTests/Dialogs/CastSizePlanningViewModelTests.cs#L446)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

How can you have a TestFixture that isn't public? Do these tests even get run? I think the class wants its own file.
</blockquote></details>

Most of our TestFixtures are not public. Yes, they run.
I agree this fixture and the other similar one deserve their own file. Thanks.

---

*[Reference Text Utility/RefTextUtilityForm.cs, line 154 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonZbzR3nEE4Sta3PKe:-Kos9CtzMAgc0wwiLb0h:b-l1u9nl) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Reference Text Utility/RefTextUtilityForm.cs#L154)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

I don't recall seeing this declared, but what's an argument doing called 'dummy'?
</blockquote></details>

I was able to work around the compile issue I was getting previously be renaming one of the overrides.
Thanks.

---

*[RefTextDevUtilities/ReferenceTextUtility.cs, line 212 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-Kon_3_IhKtgZXv2jdgY:-KosE2AnR4DEYW7tyOLQ:b-u3lgge) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/RefTextDevUtilities/ReferenceTextUtility.cs#L212)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Very confusing. For one thing it's NOT an override, so why say it is and have an extra argument deliberately not used? Needs at least a better explanation.
</blockquote></details>

Fixed another way. Thanks.

---

*[RefTextDevUtilities/ReferenceTextUtility.cs, line 214 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-Kon_KWC6GjZRZieu-Si:-Kos9gvJVi1b_q2M-ijl:b984ltp) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/RefTextDevUtilities/ReferenceTextUtility.cs#L214)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

You can now just omit the unwanted optional param
</blockquote></details>

Added comment. I think it addresses what you are saying, but I'm not sure I understand.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-:-KosL8MIGWrO-t3uvLZe:b-ed1xoc)*

JohnThomson:
Looks pretty good now. Might still be worth some discussion of the names used in the XML file and the object that serializes it. Or since this is just refactoring we could consider any such improvement later. Fine with me to go ahead and merge.

---

Review status: 40 of 78 files reviewed at latest revision, 5 unresolved discussions.

---

*[Glyssen/Block.cs, line 22 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonHnKVc2LTpO78Hvbo:-KosqDPlrT-1grNy9K_v:b-yn5gkp) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/Block.cs#L22)):*

This particular class is not being promoted. Glyssen historically hasn't bothered with class comments (not saying we shouldn't have). But I'll compromise by adding comments for classes in GlyssenShared.
</blockquote></details>

It's your code base. Probably limited usefulness to add them to just a few classes. But I find them some of the most useful of all comments. The provide context for understanding everything in the class, and clarify the intent of the author about its responsibilities, which is often crucial in deciding where to add new functionality without causing things to degenerate into spaghetti.

---

*[Glyssen/ProjectExporter.cs, line 72 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonOagQMY5-8CW2By_8:-KosrLa-s96OurSNuItC:bqv4s27) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/Glyssen/ProjectExporter.cs#L72)):*

Annotation means something very specific here. But I see your point that "Scripture" only is not accurate. I'll try to make it more clear.
</blockquote></details>

Hard to evaluate because I don't know what that very specific thing is that annotation means here. But if it's reasonably likely to be known to a maintainer of this code, I think shortening the name makes it less confusing in this case.

---

*[GlyssenShared/Script/GlyssenScript.cs, line 9 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonVBWJghq42UBtO-F5:-Kot0GA6M7CclIYtJu3l:b32w0x) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L9)):*

I agree it seems odd... but yes, it is done for serialization reasons.
I've added a comment.
</blockquote></details>

OK

---

*[GlyssenShared/Script/GlyssenScript.cs, line 82 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonXTgTPOl_9MHK4x36:-Koswos9hc2mdsAorXTI:bzcwul6) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L82)):*

Are you suggesting we may want to change the xml? Or just internally?
</blockquote></details>

Not sure.  We're free to change the xml names at this point, right, since nothing but unreleased versions of Glyssen and HearThis use it?
But once both are released, changing it will be a real pain, requiring migration code and new releases probably in both programs. So we ought to try hard to come up with the best names we can.
"ref" is unfortunately ambiguous between "reference text" (i.e., some majority language scripture version that FCBH bases things like sound effects on?) and "verse reference"; my first guess was that "primaryref" would be the starting verse and secondaryref would be the end of a bridge. "Verse" could plausibly be either the verse number or the text content of the verse, while 'vern' could as easily mean the ID of the vernacular language as the content.
There's of course some advantage to file size in having short XML names.
I don't like the reduplication in PrimaryReferenceTextText. PrimaryReferenceTextContent might be better. But again, Primary suggests the most important thing in the object, but actually it's just documentation of secondary(!) importance...HearThis doesn't even use it.
Maybe we should talk about options. Or just ignore my grumbling...you and Tom may already have worked through this enough.

---

*[RefTextDevUtilities/ReferenceTextUtility.cs, line 214 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-Kon_KWC6GjZRZieu-Si:-Kot0ASdQUoTT_CNe3aU:bekawhw) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/RefTextDevUtilities/ReferenceTextUtility.cs#L214)):*

Added comment. I think it addresses what you are saying, but I'm not sure I understand.
</blockquote></details>

I'm not sure either. What you have now looks fine.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-:-Kot0MBlHfpRDiZD5WcU:b7cj9c9)*

andrew-polk:
Review status: 40 of 78 files reviewed at latest revision, 4 unresolved discussions.

---

*[GlyssenShared/Script/GlyssenScript.cs, line 82 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonXTgTPOl_9MHK4x36:-Kovd12g8n-n_SmkeAYt:bvbqgg4) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L82)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Not sure.  We're free to change the xml names at this point, right, since nothing but unreleased versions of Glyssen and HearThis use it?
But once both are released, changing it will be a real pain, requiring migration code and new releases probably in both programs. So we ought to try hard to come up with the best names we can.
"ref" is unfortunately ambiguous between "reference text" (i.e., some majority language scripture version that FCBH bases things like sound effects on?) and "verse reference"; my first guess was that "primaryref" would be the starting verse and secondaryref would be the end of a bridge. "Verse" could plausibly be either the verse number or the text content of the verse, while 'vern' could as easily mean the ID of the vernacular language as the content.
There's of course some advantage to file size in having short XML names.
I don't like the reduplication in PrimaryReferenceTextText. PrimaryReferenceTextContent might be better. But again, Primary suggests the most important thing in the object, but actually it's just documentation of secondary(!) importance...HearThis doesn't even use it.
Maybe we should talk about options. Or just ignore my grumbling...you and Tom may already have worked through this enough.
</blockquote></details>

I agree we should definitely work through this now. Yes, changes are free for now.
I simply used what Tom proposed initially. Would definitely be worth changes if we agree they are improvements.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/glyssen/392)*

andrew-polk:
Review status: 40 of 78 files reviewed at latest revision, 4 unresolved discussions.

---

*[GlyssenShared/Script/GlyssenScript.cs, line 82 at r2](https://reviewable.io:443/reviews/sillsdev/glyssen/392#-KonXTgTPOl_9MHK4x36:-KovfBTYhk6CWy3twPSK:b-iokusv) ([raw file](https://github.com/sillsdev/glyssen/blob/d22787def24f6988f951440c16641b1c7f298c7b/GlyssenShared/Script/GlyssenScript.cs#L82)):*

I agree we should definitely work through this now. Yes, changes are free for now.
I simply used what Tom proposed initially. Would definitely be worth changes if we agree they are improvements.
</blockquote></details>

I'm going to merge this PR so I can get the other one reviewed while we have whatever discussion we want about changing the xml.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/glyssen/392)*

PR #406, opened 2017-09-21
Rework proximity calculation to use verses rather than blocks (PG-454)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/406)

### DIFF
--- a/Glyssen/Block.cs
+++ b/Glyssen/Block.cs
@@ -259,6 +259,9 @@ namespace Glyssen
 			}
 		}
 
+		// This call is not optimized. GetText() is a fairly expensive operation.
+		// A possible enhancement would be to just sum the lengths of the ScriptText elements,
+		// but that approach would need special logic for chapter announcements (see GetText).
 		public int Length => GetText(false).Length;
 
 		public int ScriptTextCount => BlockElements.Count(e => e is ScriptText);
--- a/Glyssen/Block.cs
+++ b/Glyssen/Block.cs
@@ -261,7 +261,7 @@ namespace Glyssen
 
 		public int Length => GetText(false).Length;
 
-		public int NumberOfVerses => LastVerseNum - InitialStartVerseNumber + 1;
+		public int ScriptTextCount => BlockElements.Count(e => e is ScriptText);
 
 		public string GetPrimaryReferenceText(bool textOnly = false)
 		{
--- a/Glyssen/Block.cs
+++ b/Glyssen/Block.cs
@@ -259,6 +259,10 @@ namespace Glyssen
 			}
 		}
 
+		public int Length => GetText(false).Length;
+
+		public int NumberOfVerses => LastVerseNum - InitialStartVerseNumber + 1;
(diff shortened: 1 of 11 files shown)


Glyssen/Block.cs
```diff
@@ -259,6 +259,10 @@ public bool MatchesReferenceText
 			}
 		}
 
+		public int Length => GetText(false).Length;
```
tombogle: Might want to add a comment about poor performance and/or an ENHANCE comment. I see that this is a refactoring (re-locating) of existing code, but I it looks like performance could be quite a bit better if we only counted the length of ScriptText elements. Calling GetText is fairly expensive as it has to build the composite string -- all we really need is its length. Note: ChapterAnnouncements would still need special handling if we wanted them to remain accurate.

GlyssenTests/BlockTests.cs
```diff
@@ -1375,28 +1375,113 @@ public void GetSwappedReferenceText_RowBIsNull_EntireContentsSwap()
 			Assert.AreEqual("{21} Verse twenty-one.", newRowBValue);
 		}
 
-		private CharacterVerse JesusQuestioning
+		[TestCase("1234567")]
```
tombogle: Should probably also have a couple tests for Length to cover the special case of ChapterAnnouncements.

andrew-polk:
Review status: 0 of 11 files reviewed at latest revision, 2 unresolved discussions.

---

*[Glyssen/Block.cs, line 262 at r1](https://reviewable.io:443/reviews/sillsdev/glyssen/406#-Ku_9vfSXLH2y7_k5_5x-r1-262:-Ku_W_2K5iZP5hNSFue4:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/43c8268ba27b854657b88bb99096922a1f021de0/Glyssen/Block.cs#L262)):*

Might want to add a comment about poor performance and/or an ENHANCE comment. I see that this is a refactoring (re-locating) of existing code, but I it looks like performance could be quite a bit better if we only counted the length of ScriptText elements. Calling GetText is fairly expensive as it has to build the composite string -- all we really need is its length. Note: ChapterAnnouncements would still need special handling if we wanted them to remain accurate.
</blockquote></details>

Done.

---

*[GlyssenTests/BlockTests.cs, line 1378 at r1](https://reviewable.io:443/reviews/sillsdev/glyssen/406#-Ku_9vfTlSSuSorxb8i4-r1-1378:-Ku_WaWvPeMgU8LlI41W:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/43c8268ba27b854657b88bb99096922a1f021de0/GlyssenTests/BlockTests.cs#L1378)):*

Should probably also have a couple tests for Length to cover the special case of ChapterAnnouncements.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/glyssen/406)*

PR #409, opened 2017-09-26
Fix demo disambiguation and update OT cast size numbers (PG-454)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/409)


andrew-polk:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[Glyssen/Dialogs/CastSizePlanningViewModel.cs, line 208 at r1](https://reviewable.io:443/reviews/sillsdev/glyssen/409#-Kuy9fRBXzJUwMlhiHV8:-Kuy9fRBXzJUwMlhiHV9:b4kj57a) ([raw file](https://github.com/sillsdev/glyssen/blob/d4877944c2d1f864ffb25becc8fb20b1c6b41ec1/Glyssen/Dialogs/CastSizePlanningViewModel.cs#L208)):*
> ```Smalltalk
> 						break;
> 					case "PRO":
> 						smallCast.Male = Math.Max(smallCast.Male, 1);
> ```

This one increased because when we first generated these numbers, we didn't really have PRO set up in the control file at all. Has nothing to do with this change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/glyssen/409)*

PR #448, opened 2018-06-19
Combine title and chapter announcement

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/448)

### DIFF
--- a/Glyssen/Block.cs
+++ b/Glyssen/Block.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using System.Text;
@@ -911,6 +912,27 @@ namespace Glyssen
 			else if (otherBlock.MatchesReferenceText)
 				throw new InvalidOperationException("No known need for combining blocks where only one of them is aligned to reference text.");
 			return this;
+		}
+
+		/// <summary>
+		/// This is basically a simplified version of CombineWith which assumes we are just dealing with simple text
+		/// </summary>
+		public Block AppendTextFrom(Block blockWithTextToAppend)
+		{
+			var clone = Clone(true);
+
+			Debug.Assert(clone.BlockElements.Count == 1);
+			Debug.Assert(clone.BlockElements.OfType<ScriptText>().Count() == 1);
+			Debug.Assert(blockWithTextToAppend.BlockElements.Count == 1);
+			Debug.Assert(blockWithTextToAppend.BlockElements.OfType<ScriptText>().Count() == 1);
+
+			var baseBlockText = clone.GetText(false);
+			clone.BlockElements.OfType<ScriptText>().Single().Content = baseBlockText.CombineGuaranteeingWhitespaceBetween(blockWithTextToAppend.GetText(false));
+
+			if (clone.MatchesReferenceText && blockWithTextToAppend.MatchesReferenceText)
+				clone.ReferenceBlocks[0] = clone.ReferenceBlocks.Single().AppendTextFrom(blockWithTextToAppend.ReferenceBlocks.Single());
+
+			return clone;
 		}
 
 		public bool IsFollowOnParagraphStyle => s_regexFollowOnParagraphStyles.IsMatch(StyleTag);
--- a/Glyssen/BookScript.cs
+++ b/Glyssen/BookScript.cs
@@ -8,6 +8,7 @@ using Glyssen.Character;
 using Glyssen.Dialogs;
 using Glyssen.Quote;
 using Glyssen.Shared;
+using Glyssen.Utilities;
 using SIL.Extensions;
 using SIL.Scripture;
 using SIL.Unicode;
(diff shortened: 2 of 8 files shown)


tombogle:
It would be helpful if the PR referenced the Jira issue # if there is one. I couldn't find one. In the PR, I don't see any UI changes, so I'm guessing that maybe this was a response to a request by FCBH. If I understand the change (and the intention), I wonder if we'd want to tweak the UI in Project Settings to allow the desired result. Have we considered how merging these will affect the UI in HearThis? (Currently, I think the book title is part of "Chapter 0".)

Glyssen/Block.cs
```diff
@@ -913,6 +914,27 @@ public Block CombineWith(Block otherBlock)
...
+			Debug.Assert(clone.BlockElements.Count == 1);
+			Debug.Assert(clone.BlockElements.OfType<ScriptText>().Count() == 1);
+			Debug.Assert(blockWithTextToAppend.BlockElements.Count == 1);
+			Debug.Assert(blockWithTextToAppend.BlockElements.OfType<ScriptText>().Count() == 1);
+
+			var baseBlockText = clone.GetText(false);
+			clone.BlockElements.OfType<ScriptText>().Single().Content = baseBlockText.CombineGuaranteeingWhitespaceBetween(blockWithTextToAppend.GetText(false));
+
+			if (clone.MatchesReferenceText && blockWithTextToAppend.MatchesReferenceText)
+				clone.ReferenceBlocks[0] = clone.ReferenceBlocks.Single().AppendTextFrom(blockWithTextToAppend.ReferenceBlocks.Single());
```
tombogle: Might never happen in real life, but if the clone does not have a matching reference text, but the blockWithTextToAppend does, wouldn't we probably want to clone the reference text block and stick it into clone.ReferenceBlocks?

Glyssen/ReferenceText.cs
```diff
@@ -185,6 +185,17 @@ public IEnumerable<BookScript> GetBooksWithBlocksConnectedToReferenceText(Projec
...
 					var clone = book.Clone(true);
 					ApplyTo(clone, project.Versification);
+
+					if (combineBookChapter)
+					{
+						var lastChapter = project.Versification.GetLastChapter(BCVRef.BookToNumber(book.BookId));
+						if ((!project.SkipChapterAnnouncementForFirstChapter && lastChapter > 1) ||
+						    (!project.SkipChapterAnnouncementForSingleChapterBooks && lastChapter == 1))
+						{
+							clone.Blocks = clone.CombineBookChapterBlocks();
```
tombogle: The above call to book.Clone(true) is already doing a join of other blocks, creating an entire cloned list. Now we're creating a second cloned list solely for the purpose of joining (at most) two blocks. Not sure if the performance hit is noticeable, but it seems like it would be preferable to pass a flag to that Clone method to tell it whether to do this join as well. This would be a very simple refactoring since there are only two uses of this method, and only one in production code.

Glyssen/BookScript.cs
```diff
@@ -190,6 +191,29 @@ public System.Collections.Generic.IReadOnlyList<Block> GetScriptBlocks(bool join
 			return list;
 		}
 
+		public List<Block> CombineBookChapterBlocks()
+		{
+			var list = new List<Block>(m_blockCount);
+
+			list.Add(m_blocks[0]);
+
+			for (var i = 1; i < m_blockCount; i++)
```
tombogle: Pretty sure the title block (if there is one) is always guaranteed to be the first block in the book, so this could be made much more efficient by just checking the first two blocks. We wouldn't even need to make a copy of the list if we weren't joining them.

GlyssenTests/ProjectExportTests.cs
```diff
@@ -562,11 +577,11 @@ public void GetExportData_BlocksAreJoinedToStandardNonEnglishReferenceText_Outpu
 			Assert.IsTrue(data.All(d => d.Count == 11));
 			Assert.IsTrue(data.All(d => (string)d[exporter.GetColumnIndex(ExportColumn.BookId)] == "JUD"));
 			Assert.IsTrue(data.All(d => d.Count == exporter.GetColumnIndex(ExportColumn.VernacularTextLength) + 1));
-			Assert.AreEqual("Иуда", data[0][exporter.GetColumnIndex(ExportColumn.PrimaryReferenceText)]);
-			Assert.AreEqual("JUDE", data[0][exporter.GetColumnIndex(ExportColumn.SecondaryReferenceText)]);
+			Assert.AreEqual("Иуда Иуда 1", data[0][exporter.GetColumnIndex(ExportColumn.PrimaryReferenceText)]);
```
tombogle: This test case illustrates what seems to be the potential fallacy in doing this at all. With no punctuation or anything to help the voice actor, this seems quite prone to being misread. (Not necessarily a problem for the FCBH teams with an experience coach, but likely to trip up SIL teams doing self-recording.)

andrew-polk:
After nearly 2 years, I think it's probably time to close this one. 
Too many questions. 
And too many merge conflicts now. : )

PR #452, opened 2018-07-06
Allow user to split a block which has leading punctuation (PG-1107)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/452)

### DIFF
--- a/Glyssen/Block.cs
+++ b/Glyssen/Block.cs
@@ -30,6 +30,9 @@ namespace Glyssen
 						".right-to-left{{direction:rtl}}" +
 						".scripttext {{display:inline}}";
 
+		public const string kLeadingPunctuationHtmlStart = "<span class=\"leading-punctuation\">";
+		public const string kLeadingPunctuationHtmlEnd = "</span>";
+
 		private static readonly Regex s_regexFollowOnParagraphStyles;
 		internal static Regex s_regexInterruption;
 
@@ -545,18 +548,16 @@ namespace Glyssen
 			return bldr.ToString();
 		}
 
-		public bool TryGetLeadingPunctuation(out ScriptText scriptTextWithLeadingPunctuation)
+		private string GetLeadingPunctuation()
 		{
 			if (BlockElements.FirstOrDefault() is ScriptText initialScriptText &&
 			    BlockElements.Skip(1).FirstOrDefault() is Verse initialVerse &&
 			    initialVerse.Number == InitialVerseNumberOrBridge)
 			{
-				scriptTextWithLeadingPunctuation = initialScriptText;
-				return true;
+				return initialScriptText.Content;
 			}
 
-			scriptTextWithLeadingPunctuation = null;
-			return false;
+			return null;
 		}
 
 		public string GetSplitTextAsHtml(int blockId, bool rightToLeftScript, IEnumerable<BlockSplitData> blockSplits, bool showCharacters)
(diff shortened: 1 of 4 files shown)


Glyssen/Block.cs
```diff
@@ -545,18 +545,41 @@ public string GetTextAsHtml(bool showVerseNumbers, bool rightToLeftScript)
 			return bldr.ToString();
 		}
 
+		public bool TryGetLeadingPunctuation(out ScriptText scriptTextWithLeadingPunctuation)
+		{
```
tombogle: Given this name, I think I would expect it to return the text itself (i.e., return initialScriptText.Content) rather than the ScriptText.

Glyssen/Block.cs
```diff
@@ -545,18 +545,41 @@ public string GetTextAsHtml(bool showVerseNumbers, bool rightToLeftScript)
...
+
+			// Look for special case where verse has leading punctuation before the verse number such as
+			// ({1} This verse is surrounded by parentheses)
+			// This can only happen at the beginning of a block.
+			// If we have it, we basically want to do the split as if it wasn't there at all. i.e. Split the main part of the verse only, and
+			// do not include the leading punctuation as part of the offset.
+			if (TryGetLeadingPunctuation(out ScriptText initialScriptText))
+				leadingPunctuationHtml = Format(leadingPunctuationTemplate, initialScriptText.Content);
+
+			foreach (var blockElement in BlockElements.Skip(leadingPunctuationHtml != Empty ? 1 : 0))
```
tombogle: Interesting that you use a TryGet pattern here, apparently to avoid the cumbersomeness of having to check for a null return value, but then right below you have to do a check for Empty (or null if you make the change I suggest above).

Glyssen/Block.cs
```diff
@@ -545,18 +545,41 @@ public string GetTextAsHtml(bool showVerseNumbers, bool rightToLeftScript)
...
+
 		public string GetSplitTextAsHtml(int blockId, bool rightToLeftScript, IEnumerable<BlockSplitData> blockSplits, bool showCharacters)
 		{
 			var bldr = new StringBuilder();
 			var currVerse = InitialVerseNumberOrBridge;
 			var verseNumberHtml = Empty;
-			const string splitTextTemplate = "<div class=\"splittext\" data-blockid=\"{2}\" data-verse=\"{3}\">{0}{1}</div>";
-
-			foreach (var blockElement in BlockElements)
+			var leadingPunctuationHtml = Empty;
```
tombogle: Using a null here would be fine. String.Format handles nulls no problem.

Glyssen/Dialogs/SplitBlockDlg.cs
```diff
@@ -24,7 +24,7 @@ public partial class SplitBlockDlg : FormWithPersistedSettings
 		private int m_blockSplitIdCounter = 1;  // zero is reserved for assigning a character id to the first segment
 		private readonly IEnumerable<AssignCharacterViewModel.Character> m_characters;
 		private readonly string m_css = Resources.BlockSplitCss;
-		private readonly Regex m_verseNumberRegex = new Regex(@"<sup>.*</sup>");
+		private readonly Regex m_verseNumberRegex = new Regex(@"(<span.*?>.+?<\/span>)+<sup>.*?<\/sup>");
```
tombogle: I'm not clear as to what motivated this change. It looks like you're saying that any (potentially nested) spans immediately preceding the super-scripted text should also be considered part of the verse number. But why?

Glyssen/Dialogs/SplitBlockDlg.cs
```diff
@@ -24,7 +24,13 @@ public partial class SplitBlockDlg : FormWithPersistedSettings
...
 		private readonly IEnumerable<AssignCharacterViewModel.Character> m_characters;
 		private readonly string m_css = Resources.BlockSplitCss;
-		private readonly Regex m_verseNumberRegex = new Regex(@"(<span.*?>.+?<\/span>)+<sup>.*?<\/sup>");
+
+		// For purposes of splitting, anyway, we treat leading punctuation as if is permanently attached to the
+		// beginning of the verse just as much as the verse number is. So the user may not split within
+		// (or just after) the leading punctuation. This means the split index is always without regard to
+		// the presence or absence of leading punctuation.
+		// This was perhaps more of an ease-of-implementation decision rather than a decision made on its merits.
+		private readonly Regex m_verseNumberRegex = new Regex(@"(" + Regex.Escape(Block.kLeadingPunctuationHtmlStart) + ".*?" + Regex.Escape(Block.kLeadingPunctuationHtmlEnd) + ")+<sup>.*?</sup>");
```
tombogle: This is much more readable. However, I would have expected a ? instead of a + for the leading punctuation stuff. There can't be more than one, can there?

PR #460, opened 2018-10-26
PG-63 Added support for accessing data from live Paratext Projects

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/460)


tombogle:
___
*[DistFiles/localization/Glyssen.es.tmx, line 4 at r1](https://reviewable.io/reviews/sillsdev/glyssen/460#-LREJ_276AnG6rCTvPZg:-LRNDW6M1m025AjSGTPZ:bxjkpz) ([raw file](https://github.com/sillsdev/glyssen/blob/b58bdbb570e707c95f00be6037af3c7378c18612/DistFiles/localization/Glyssen.es.tmx#L4)):*

Might as well go ahead and make it 1.1 if we are changing it.
</blockquote></details>

That was generated. The next generated version (with some newly translated strings) will have the new version #.

tombogle:
___
*[Glyssen/DataMigrator.cs, line 102 at r1](https://reviewable.io/reviews/sillsdev/glyssen/460#-LREKg4M-pCnXwc68_NP:-LRNLZAu0j9mMv1nlwsf:bwg4obs) ([raw file](https://github.com/sillsdev/glyssen/blob/b58bdbb570e707c95f00be6037af3c7378c18612/Glyssen/DataMigrator.cs#L102)):*

Isn't this for cleaning up after migrating the project data down one level or something?
Like, didn't we add a level at one point?
</blockquote></details>

This migration step is to add missing versification files (which we didn't used to save in  version 0). The point o f this comment is that in later versions (3+), it could be valid for OriginalReleaseBundlePath to be null or empty because the project could be Paratext-based, in which case there is no original bundle path. But in versions 0 and 1, if that metadata wasn't set the project was corrupt/bogus.

tombogle:
If the user ever overrides a failing check and adds a book, the project metadata will contain a flag (per book) indicating that. At the time they make the decision, I also write out an entry in the log file to that effect, so it should be clear if the failure happens during the parsing stage (which is the most likely place for us to fall down). If they get a failure further down the road after quitting and restarting Glyssen, it will be less obvious. Even if we see that flag in the project metadata, it probably won't be immediately clear whether the failure was actually a result of bad data.

tombogle:
___
*[Glyssen/MainForm.cs, line 878 at r1](https://reviewable.io/reviews/sillsdev/glyssen/460#-LREMU4x9oTVDBzpXN1Q:-LRNOX2s6cVKJnyuONMn:b-lmz5kc) ([raw file](https://github.com/sillsdev/glyssen/blob/b58bdbb570e707c95f00be6037af3c7378c18612/Glyssen/MainForm.cs#L878)):*
> ```
> private void SelectBooks_Click(object sender, EventArgs e)
> ```
Nope. Fixed the indentation.

tombogle:
___
*[Glyssen/Project.cs, line 2300 at r1](https://reviewable.io/reviews/sillsdev/glyssen/460#-LREQkytB7_WEDvAviop:-LRNQbpTEyT6wJVCShD9:bnmcvdz) ([raw file](https://github.com/sillsdev/glyssen/blob/b58bdbb570e707c95f00be6037af3c7378c18612/Glyssen/Project.cs#L2300)):*

> ```
> Cannot text quote system.
> ```
Not sure what you intended to say, but I'm pretty sure this wasn't it. : )conv
</blockquote></details>

"test" Fixed.

tombogle:
___
*[Glyssen/Controls/ExistingProjectsList.cs, line 37 at r4](https://reviewable.io/reviews/sillsdev/glyssen/460#-LRIVCsm0yjbP3_hp74U:-LRNRgOa24V_W3UYc_Sf:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/5d0b1ac5a09caff124c072445ea3584ba4938bf2/Glyssen/Controls/ExistingProjectsList.cs#L37)):*

Wouldn't be a bad idea to include a comment here regarding what {0} is as you have in other places.
</blockquote></details>

Done.

tombogle:
___
*[Glyssen/Dialogs/OpenProjectDlg.cs, line 100 at r4](https://reviewable.io/reviews/sillsdev/glyssen/460#-LRIajNR6LAuJsoK5eRl:-LRNSBGBEQQ5oUBSaOec:b-3181hx) ([raw file](https://github.com/sillsdev/glyssen/blob/5d0b1ac5a09caff124c072445ea3584ba4938bf2/Glyssen/Dialogs/OpenProjectDlg.cs#L100)):*

> ```
> HearThis
> ```
Glyssen
Another a few lines down.
</blockquote></details>

Done. Caught those mistakes yesterday :-)

tombogle:
___
*[Glyssen/Dialogs/ScriptureRangeSelectionDlg.cs, line 426 at r4](https://reviewable.io/reviews/sillsdev/glyssen/460#-LRI_NRM3TJ_eN1qzOUL:-LRNSPg2ByweWZ1eFtva:b-v3emn6) ([raw file](https://github.com/sillsdev/glyssen/blob/5d0b1ac5a09caff124c072445ea3584ba4938bf2/Glyssen/Dialogs/ScriptureRangeSelectionDlg.cs#L426)):*
> IsValidToIncludeBook
No, it only gets called if they are trying to select it, not if they are trying to clear it. (See conditional checks for e.FormattedValue and grid[e.ColumnIndex, e.RowIndex].EditedFormattedValue.)

andrew-polk:
But let's remove WIP from the PR summary

PR #465, opened 2018-12-17
PG-1146: Made Glyssen get the quotation rules from Paratext…

tombogle:
…settings instead of the LDML fi;le, which might be out-of-date.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/465)

### DIFF
--- a/Glyssen/Dialogs/ProjectSettingsDlg.cs
+++ b/Glyssen/Dialogs/ProjectSettingsDlg.cs
@@ -85,7 +85,7 @@ namespace Glyssen.Dialogs
 			var project = m_model.Project;
 			if (IsHandleCreated && Visible && project.IsLiveParatextProject &&
 				(project.QuoteSystemStatus & QuoteSystemStatus.ParseReady) != 0 &&
-				project.WritingSystem.QuotationMarks.SequenceEqual(project.GetLiveParatextDataIfCompatible(false, "", false)?.QuotationMarks ?? new QuotationMark[] { }))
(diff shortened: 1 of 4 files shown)


tombogle:
Yes

tombogle:
___
*[Glyssen/Dialogs/ProjectSettingsDlg.cs, line 88 at r1](https://reviewable.io/reviews/sillsdev/glyssen/465#-LTxA6jfB5_Jg2c7Bhif:-LTxg_3OEtN8Gh6A5AsE:brhx1u0) ([raw file](https://github.com/sillsdev/glyssen/blob/110c7dfcf71fc436f9e814d7ce6da4923ce91ce1/Glyssen/Dialogs/ProjectSettingsDlg.cs#L88)):*

So this is a bug fix in its own right, right? (Apart from the LDML issue?)
</blockquote></details>

Yes. (This is actually the bug I was trying to track down when I discovered the other proble,.)

tombogle:
___
*[Glyssen/Paratext/ParatextScrTextWrapper.cs, line 58 at r1](https://reviewable.io/reviews/sillsdev/glyssen/465#-LTxAoYCCNkgDdCbjoOK:-LTxgno3-80lPVAWB7w6:btzyqrj) ([raw file](https://github.com/sillsdev/glyssen/blob/110c7dfcf71fc436f9e814d7ce6da4923ce91ce1/Glyssen/Paratext/ParatextScrTextWrapper.cs#L58)):*

Here is a question to show my ignorance...
If the PT text changes, we don't get those changes unless the Glyssen user says they want them, right? So should we only get the latest quote mark settings in the same way?
i.e. should we just read the PT settings into our LDML when we get the text and read them the same way as other projects?
</blockquote></details>

We will only get the updated Paratext project settings if the user clicks the Update button to update the data.

PR #469, opened 2019-02-12
PG-1123, PG-1131: Fixed "Sequence contains no matching element"

tombogle:
This is the case of this crash that happens when moving a ref block whose character is not set.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/469)

### DIFF
--- a/Glyssen/Dialogs/AssignCharacterDlg.cs
+++ b/Glyssen/Dialogs/AssignCharacterDlg.cs
@@ -1459,26 +1459,35 @@ namespace Glyssen.Dialogs
 					if (selectedCharacter == null)
 					{
 						var newValue = m_dataGridReferenceText.Rows[e.RowIndex].Cells[e.ColumnIndex].Value as string;
-						selectedCharacter = colCharacter.Items.Cast<AssignCharacterViewModel.Character>()
-							.First(c => c.LocalizedDisplay == newValue);
+						if (newValue != null)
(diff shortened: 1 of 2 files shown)


tombogle:
___
*[Glyssen/Dialogs/AssignCharacterDlg.cs, line 1469 at r1](https://reviewable.io/reviews/sillsdev/glyssen/469#-LYYpvREE-fG3WemFwP5:-LYZnSxj6MKi30X_cvr1:bwd86m8) ([raw file](https://github.com/sillsdev/glyssen/blob/317f9d6255ae63dd99ff821e3d6f191800da636c/Glyssen/Dialogs/AssignCharacterDlg.cs#L1469)):*

> ```
> m_viewModel.SetReferenceTextMatchupCharacter(e.RowIndex, selectedCharacter);
> ```

So `m_viewModel.SetReferenceTextMatchupCharacter(e.RowIndex, null);` sets it to Ambiguous?
And selectedCharacter remaining null throughout this code is okay?
</blockquote></details>

Yes. I walked through the code and then tested it a few times using various similar scenarios. I'm still not 100% sure this is the *best* fix. (An alternative might be to try to prevent swapping rows unless they both have a character set. Or maybe force the blank one to narrator.) I tried to imagine how an FCBH scripter would use this in real life, but I think the final proof will be in having them actually use it. Anyway, by preventing the crash, I think we give them a way forward.

PR #472, opened 2019-02-19
PG-1162: Updated Glyssen's OT reference text (English) based on new version of Director's Guide from FCBH

tombogle:
This includes a large number of enhancements to Glyssen to handle new requirements and situations pertinent to the Old Testament, notably:
- PG-40: Make use of Implicit quote type, properly assigning the specified character in the quote parser unless the verse(s) have explicit quotes. Related to that, added Alternate quote type and made quote parser able to disregard it, except when needed to cover a quote by a character that continues into following blocks for verses where that character is only an Alternate.

- PG-1198: Allow Scripter to assign "Needs Review" to a block instead of an actual biblical character.

- PG-1201: Allow hypothetical characters to display, but only assign them by default if they are used in the reference text

- PG-1202: Eliminated code to make some OT books single-voice by default

- Added capability to apply narrator overrides to allow specific Psalms to be assigned to their authors and to allow other OT books to deal with the switching back and forth between 3rd-person and 1st-person narration.

- Made improvements to the dev tool for processing the guide.

- Enhanced MigrateDeprecatedCharacterIds to work properly with reference text blocks created by processing the director's guide.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/472)


tombogle:
___
*[DevTools/CharacterDetailProcessing.cs, line 32 at r2](https://reviewable.io/reviews/sillsdev/glyssen/472#-LmzXQzv9vNZkSpUB2rv:-LnD42F18CHQTC7GfMW3:b-dk14vi) ([raw file](https://github.com/sillsdev/glyssen/blob/9e06601b1fda9adfbed901ec0c777fb2ee98c532/DevTools/CharacterDetailProcessing.cs#L32)):*
> dictionaryWithHypotheticals
Unless you specify the second argument to limit the number of splits, it will create an array with one item per slash + 1. The data integrity test should catch the edge case of two slashes together.

tombogle:
___
*[Glyssen/Character/CharacterVerse.cs, line 160 at r3](https://reviewable.io/reviews/sillsdev/glyssen/472#-Ln-EI7c5lvp2f-GAJBD:-LnOzXQy5ohqHN2W96Gd:b-y0wqe8) ([raw file](https://github.com/sillsdev/glyssen/blob/4d4ff0ec8a65c123c341d52b72742fcfff6873ab/Glyssen/Character/CharacterVerse.cs#L160)):*

Sorry; I'm no help there. I know we had the Scripture thing where we tried to capture the original speaker. Beyond that, you would have to see if/how else we use it in the control file.
</blockquote></details>

I used this RegEx to find them (excluding the scripture Quotations): ^\w*\t\d*\t\d*\t[^\t/s][^\t/c]?[^\t/r]?[^\t/i]*\t[^\t]*\t[^\t]*\t[^\t]*\t\w[^\t]*\t
The other places we use it: 1) to note that a character is "Good Priest"; 2) to translate from generic groups (i.e., "parents" or "disciples", "people in high priest's courtyard") to specific individuals ("father", "mother", "Thomas", etc.); 3) to indicate a guess as to who the real "voice from heaven" is; 4) to translate possible OT theophanies to Jesus; 5) entries in JHN 21:7 and JHN 21:17 that sit alongside the "correct" entry that basically tell the user to have the narrator read it; 6) In GEN 28:6, there's an entry to attempt to say that a quotation by Isaac should be read by the narrator (even though FCBH expects Isaac to say it); 7) Some places where we have a narrator Quotation (e.g., PRO 30:8), we note that it should be read by Agur, but now this is actually handled by narrator overrides; 8) dream => angel.
Probably some of them should be handled differently...

tombogle:
___
*[Glyssen/Resources/CharacterDetail.txt, line 1 at r6](https://reviewable.io/reviews/sillsdev/glyssen/472#-LnJHCVH5Zk9B9GHNFUK:-LnswTTmCdYESxzLVNmY:bb483es) ([raw file](https://github.com/sillsdev/glyssen/blob/65282cee13520d036107a44617f53318189cbd6e/Glyssen/Resources/CharacterDetail.txt#L1)):*
> he general case is the quote parser ignore alternate characters, but it if it trying to resolve a multi-block quote, we want to include them in case an alternate is the one being continued from a previous block?
probably not worth reviewing. The only changes likely to be even remotely interesting are the handful of places where I renamed or split a character.

tombogle:
___
*[Glyssen/Character/NarratorOverrides.cs, line 156 at r10](https://reviewable.io/reviews/sillsdev/glyssen/472#-Lnt3DB13KXIZAgDuQ_x:-LoW_zsV1YZJNTmcR0zm:b-vgtdf0) ([raw file](https://github.com/sillsdev/glyssen/blob/3811d1b8d81a28106426125090fecfb4c8ddae35/Glyssen/Character/NarratorOverrides.cs#L156)):*
> ```
> typically property
> ```
Got my mords wixed up. Should say: property typically

tombogle:
___
*[DistFiles/reference_texts/English/1CH.xml, line 1 at r19](https://reviewable.io/reviews/sillsdev/glyssen/472#-LofJti72VtX3yHuPWIl:-LpoCE-Q9pssWAoo31HM:b-eo06lt) ([raw file](https://github.com/sillsdev/glyssen/blob/d5176e79d52cd941a9cb2ed83a7d3676b8691f21/DistFiles/reference_texts/English/1CH.xml#L1)):*

Of course, you'll need to regenerate the Russian, too.
</blockquote></details>

There is no Russian OT Director's Guide or reference text yet.

tombogle:
___
*[DistFiles/reference_texts/English/2CH.xml, line 591 at r20](https://reviewable.io/reviews/sillsdev/glyssen/472#-Logz1JkB16UWThq_77J:-LpoCUFx9wXqQmntOpXF:bq1u83m) ([raw file](https://github.com/sillsdev/glyssen/blob/4789e68ce47aadefa8c5950c6313aeed8ed8a8de/DistFiles/reference_texts/English/2CH.xml#L591)):*

I thought the idea was to make the RT match the DG.
</blockquote></details>

There are four places where my analysis of the text led me to conclude that, despite a definite decision in the DG by FCBH, it would be nearly impossible for a scripter to reach a decision without knowledge of the target language. I had already called two of these to FCBH's attention. I have now added Trello cards for the other two. Three of the four involve ambiguity where the gender of the speaker is in question. Here in 1CH 7:6, although all the choices are male, there are three viable options, but choosing the wrong option would likely sound wrong or confusing.

tombogle:
___
*[Glyssen/Character/CombinedCharacterVerseData.cs, line 50 at r21](https://reviewable.io/reviews/sillsdev/glyssen/472#-Lpefd9zEXdKYcS6rk5t:-LpoML-O1Qh72neglvqZ:b-bvd6os) ([raw file](https://github.com/sillsdev/glyssen/blob/d6b25bdc700119ae6e52abe1dedb028db3609169/Glyssen/Character/CombinedCharacterVerseData.cs#L50)):*

This is a nice optimization. Saving the union wasn't worth it for the following three methods?
</blockquote></details>

This one is different in that the "result" we start off with is the real full set (which is why it has to be IReadOnlySet, as opposed to ISet). The other ones use UnionWith, which actually modifies the set and is presumably extremely efficient in the case where the project set is empty. However in this case, unless we check for the common case of an empty project set, we don't know whether we need to make a new set (based on the Union, which just produces an IEnumerable, not a set) or not.

tombogle:
___
*[Glyssen/Quote/QuoteParser.cs, line 650 at r10](https://reviewable.io/reviews/sillsdev/glyssen/472#-Lnt5ZAm6oCrmw0-xeV9:-LpoOqu9DRx5HNfm-MXb:b-pr2has) ([raw file](https://github.com/sillsdev/glyssen/blob/3811d1b8d81a28106426125090fecfb4c8ddae35/Glyssen/Quote/QuoteParser.cs#L650)):*
> one entry with a matching character
Then apparently you aren't fluent in my dialect of hick.

tombogle:
___
*[Glyssen/Quote/QuoteParser.cs, line 650 at r10](https://reviewable.io/reviews/sillsdev/glyssen/472#-Lnt5ZAm6oCrmw0-xeV9:-LpoPQJV1fyUotRZQtyN:bhd100z) ([raw file](https://github.com/sillsdev/glyssen/blob/3811d1b8d81a28106426125090fecfb4c8ddae35/Glyssen/Quote/QuoteParser.cs#L650)):*

> one entry with a matching character
Then apparently you aren't fluent in my dialect of hick.
</blockquote></details>

So I think the right way to say what I was trying to say was "There will almost ever only be one entry..." But your way sounds more gooder.

tombogle:
___
*[RefTextDevUtilities/ReferenceTextUtility.cs, line 977 at r18](https://reviewable.io/reviews/sillsdev/glyssen/472#-LoXAcSN-D1UuDIMJT_r:-LpoRjdE56vGHK0X3QKn:b-dp0mwt) ([raw file](https://github.com/sillsdev/glyssen/blob/593aeed0a1092c370194d0b85c6066507ffb3abe/RefTextDevUtilities/ReferenceTextUtility.cs#L977)):*

You've changed it to SingleOrDefault. So where would we expect this exception to come from now?
</blockquote></details>

SingleOrDefault throws an InvalidOperationException if there is more than one.

PR #476, opened 2019-02-27
PG-1121: Fixed quote parser bug...

tombogle:
where previous block was left as start block for a quote introduced by a colon

Also stuck in a minor internationalization bug fix.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/476)

### DIFF
--- a/Glyssen/App.config
+++ b/Glyssen/App.config
@@ -70,7 +70,7 @@
     <applicationSettings>
         <Glyssen.Properties.Settings>
             <setting name="ParserVersion" serializeAs="String">
-                <value>42</value>
+                <value>43</value>
             </setting>
             <setting name="DataFormatVersion" serializeAs="String">
                 <value>3</value>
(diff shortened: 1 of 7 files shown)


tombogle:
___
*[Glyssen/Resources/CharacterVerse.txt, line 16612 at r1](https://reviewable.io/reviews/sillsdev/glyssen/476#-LZlTbbB9I2UROnFwWrA:-LZooKPQ3T9ObZM8tSTo:b-n3o2bk) ([raw file](https://github.com/sillsdev/glyssen/blob/9f3390254467082ee2fd14f246972512aaee0cde/Glyssen/Resources/CharacterVerse.txt#L16612)):*

Should this correspond to a version number bump for this file, too?
Otherwise, :lgtm: 
</blockquote></details>

In theory, maybe. I was thinking that I had already bumped up the number for another (as yet unpublished) change, but that's PG-1162, and it's not likely to get published any time soon. As it turns out, because this is a very minor addition of an unlikely character, bumping up the version would only affect existing Glyssen projects that have someone speaking in this verse but with have not already been manually disambiguated. If they did add this character here, the next time SaveProjectCharacterVerseData gets called, Glyssen will notice that it's now in the control file, and it will be removed from the project-specific one. So I think that it's not necessary to bump up the number (and thus avoid possibly forgetting to re-bump it up for PG-1162 when we merge that branch).

PR #477, opened 2019-03-01
PG-1166: Fixed off-by-one error, going off the end of a block matchup

tombogle:
Also prevent creating a pending blank character for an unassigned block from the split dialog

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/477)


tombogle:
Changing the <= to < on line 490 is what fixed the off-by-1 error.

andrew-polk:
Yup. Totally missed that change...

PR #482, opened 2019-03-05
PG-1168: Improved logic for re-applying user decisions to handle reference text alignments

tombogle:
[WIP]: Added failing unit test to illustrate bug

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/482)

### DIFF
--- a/Glyssen/Block.cs
+++ b/Glyssen/Block.cs
@@ -1275,6 +1275,8 @@ namespace Glyssen
 
 	public class BlockComparer : IEqualityComparer<Block>
 	{
+		readonly BlockReferenceComparer m_referenceComparer = new BlockReferenceComparer();
+
 		public bool Equals(Block x, Block y)
 		{
 			if (x == null && y == null)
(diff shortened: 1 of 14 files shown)


tombogle:
___
*[Glyssen/BlockMatchup.cs, line 229 at r1](https://reviewable.io/reviews/sillsdev/glyssen/482#-LapijmH1C81BIKvXb6l:-LauDw1w-FiPxVw77lzO:bjo78wh) ([raw file](https://github.com/sillsdev/glyssen/blob/9f298dd0b041eea87a125e696204e1ab5e1e2a2b/Glyssen/BlockMatchup.cs#L229)):*

Perhaps this should be InvalidOperationException, too?
</blockquote></details>

In one sense that would be ideal, but I mainly put it in for my benefit while working on the implementation and so that any future development that broke the contract would hopefully get caught by the developer. It's not necessarily a trivial check and one of the code paths for this method has it getting called repeatedly in a loop, so I didn't want to slow it down in production, especially since there's at least a fighting chance Glyssen might be able to survive the bad data condition.

tombogle:
___
*[Glyssen/Project.cs, line 1127 at r1](https://reviewable.io/reviews/sillsdev/glyssen/482#-Lapge2hCuGK2dSc-I1A:-LauFsAf0ILiaM9deXdj:b-jn7iyx) ([raw file](https://github.com/sillsdev/glyssen/blob/9f298dd0b041eea87a125e696204e1ab5e1e2a2b/Glyssen/Project.cs#L1127)):*
> ```
> UpgradedProjectOnQuoteParseCompleted
> ```
I suppose, given that I can't remember what I added it for and I'm not even sure how to parse the name...

PR #485, opened 2019-03-25
PG-1140: Disregard a "cl" marker that does not occur before or immediately after chapter

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/485)


tombogle:
___
*[Glyssen/UsxParser.cs, line 273 at r1](https://reviewable.io/reviews/sillsdev/glyssen/485#-LaqCWtA6BXpbduOMecC:-Lau3dpF8JOLvp9mqMBe:blingyn) ([raw file](https://github.com/sillsdev/glyssen/blob/92c8ee9839204b1513cc3f3118a12a716b2d065c/Glyssen/UsxParser.cs#L273)):*

Doesn't this logic rather mean that if we have no `c` markers but do find a `cl` marker, we apply the `cl` marker (regardless of where we find it)? 
Or am I just confused?
</blockquote></details>

Assuming I understand your question, I think you're just confused. There should always be one or more \c markers in a book. Otherwise, it should fail the markers check, because virutally all markers fall hierarchically under the \c. A \cl before the first \c means that in this book, the word in the \cl line should be used as the word "chapter". Most commonly, this is used in the book of Psalms to make it say Psalm 1, Psalm 2, etc. But it is possible to use this same approach in all the books so that the chapters will be labeled as Matthew 1, Matthew 2, Mark 1, etc. (This could perhaps be useful in a cultural setting where there is no meaningful concept of a chapter.)

tombogle:
___
*[Glyssen/UsxParser.cs, line 276 at r1](https://reviewable.io/reviews/sillsdev/glyssen/485#-LaqCsq31xfXBkXRotzk:-Lau4avCAbh8R1MAq-uR:bh26dk6) ([raw file](https://github.com/sillsdev/glyssen/blob/92c8ee9839204b1513cc3f3118a12a716b2d065c/Glyssen/UsxParser.cs#L276)):*

How would the last block be a chapter announcement?
I must really be missing something...
</blockquote></details>

The normal situation when a \cl is not used to set the chapter label for the whole book is to put it immediately after the \c. This is seldom used outside of the Apocrypha. It is needed in the Apocrypha, because there is a book that has some chapters marked with letters instead of numbers. When used after the \c marker, the \cl marker is the presentation form, while the \c is still needed to allow programs to properly identify chapter breaks. Outside of the Apocrypha, this usage would also make it possible to have chapter 1 be labelled "First Chapter", chapter 2 be "Second Chapter", etc.

tombogle:
___
*[GlyssenTests/UsxParserTests.cs, line 307 at r1](https://reviewable.io/reviews/sillsdev/glyssen/485#-LaqAZf49CN1s237clBH:-LauCerDDWhTy7az0Hr6:b-mpyt29) ([raw file](https://github.com/sillsdev/glyssen/blob/92c8ee9839204b1513cc3f3118a12a716b2d065c/GlyssenTests/UsxParserTests.cs#L307)):*

Would it be a better test to have this not be the last thing in the chapter? 
</blockquote></details>

I made it the last thing in the chapter because that's the way it was in the actual bad data that prompted this. I guess I could add another test, but in light of the actual implementation, it really shouldn't matter.

PR #490, opened 2019-04-15
PG-792: Improved USX parser to prevent verses with only punctuation

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/490)

### DIFF
--- a/Glyssen/App.config
+++ b/Glyssen/App.config
@@ -70,7 +70,7 @@
     <applicationSettings>
         <Glyssen.Properties.Settings>
             <setting name="ParserVersion" serializeAs="String">
-                <value>43</value>
+                <value>44</value>
             </setting>
             <setting name="DataFormatVersion" serializeAs="String">
                 <value>3</value>
--- a/Glyssen/Project.cs
+++ b/Glyssen/Project.cs
@@ -1399,7 +1399,7 @@ namespace Glyssen
 		internal void IncludeBooksFromParatext(ParatextScrTextWrapper wrapper, ISet<int> bookNumbers,
 			Action<BookScript> postParseAction)
 		{
-			wrapper.IncludeBooks(new HashSet<string>(bookNumbers.Select(BCVRef.NumberToBookCode)));
+			wrapper.IncludeBooks(bookNumbers.Select(BCVRef.NumberToBookCode));
 			var usxBookInfoList = wrapper.GetUsxDocumentsForIncludedParatextBooks(bookNumbers);
 
 			void EnhancedPostParseAction(BookScript book)
--- a/Glyssen/Properties/Settings.Designer.cs
+++ b/Glyssen/Properties/Settings.Designer.cs
@@ -73,7 +73,7 @@ namespace Glyssen.Properties {
         
         [global::System.Configuration.ApplicationScopedSettingAttribute()]
         [global::System.Diagnostics.DebuggerNonUserCodeAttribute()]
-        [global::System.Configuration.DefaultSettingValueAttribute("43")]
+        [global::System.Configuration.DefaultSettingValueAttribute("44")]
         public int ParserVersion {
             get {
                 return ((int)(this["ParserVersion"]));
--- a/Glyssen/Properties/Settings.settings
+++ b/Glyssen/Properties/Settings.settings
@@ -15,7 +15,7 @@
       <Value Profile="(Default)" />
     </Setting>
     <Setting Name="ParserVersion" Type="System.Int32" Scope="Application">
-      <Value Profile="(Default)">43</Value>
+      <Value Profile="(Default)">44</Value>
     </Setting>
     <Setting Name="DefaultSfmDirectory" Type="System.String" Scope="User">
       <Value Profile="(Default)" />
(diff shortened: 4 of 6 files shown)


tombogle:
___
*[Glyssen/UsxParser.cs, line 237 at r1](https://reviewable.io/reviews/sillsdev/glyssen/490#-LcY3uHs7kuv6_S60ZNF:-LcaGe3l2LslSK6c_uZp:bz6hdgm) ([raw file](https://github.com/sillsdev/glyssen/blob/6ab7ae793f96a2cca0e5ff587513baae95ce1878/Glyssen/UsxParser.cs#L237)):*

> ```
> block.BlockElements.Count > 1
> ```
Can the user insert a paragraph marker mid verse?
</blockquote></details>

Yes. Without this check any leading punctuation, such as an opening bracket to surround a verse with weak manuscript evidence, gets clobbered unconditionally. I'm not 100% sure, but I think it's also possible that a dialogue dash or an opening quote mark could sometimes precede the verse number as well. (I don't know if we have any tests for that case or if it is actually properly supported, but I do know we have tests for an opening bracket, and they fail without this check.)

tombogle:
___
*[Glyssen/UsxParser.cs, line 240 at r1](https://reviewable.io/reviews/sillsdev/glyssen/490#-LcY4oKW2zCGSDESj14E:-LcaJzq84_PaiVHzSCdL:bjskwhd) ([raw file](https://github.com/sillsdev/glyssen/blob/6ab7ae793f96a2cca0e5ff587513baae95ce1878/Glyssen/UsxParser.cs#L240)):*

This confused me for a bit. Probably worth a comment that this is for removing the Verse element.
</blockquote></details>

Done. I re-worked the logic and added some comments. Maybe it'll make more sense now.

tombogle:
___
*[GlyssenTests/UsxParserTests.cs, line 648 at r1](https://reviewable.io/reviews/sillsdev/glyssen/490#-LcY2PgUDOcT9uxCmnRM:-LcaKDvY3mlzCh1uhurh:bg30yk2) ([raw file](https://github.com/sillsdev/glyssen/blob/f3460fe8656d78d7841cfa4ed925fff9f8f047a3/GlyssenTests/UsxParserTests.cs#L648)):*

Seems like this comment could have been helpful, but ok...
</blockquote></details>

It was a paste-o from the above test. It didn't actually apply to this test.

tombogle:
___
*[Glyssen/UsxParser.cs, line 226 at r1](https://reviewable.io/reviews/sillsdev/glyssen/490#-LcY3KF8DpF2PsSSryeq:-LcaGbcc0-QDsJNXCJDS:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/6ab7ae793f96a2cca0e5ff587513baae95ce1878/Glyssen/UsxParser.cs#L226)):*

Probably worth a comment to inform as to what the return values means.
</blockquote></details>

Done.

tombogle:
___
*[Glyssen/UsxParser.cs, line 237 at r1](https://reviewable.io/reviews/sillsdev/glyssen/490#-LcY3uHs7kuv6_S60ZNF:-LcaKTNg1IB11H6gWnxK:b8u641n) ([raw file](https://github.com/sillsdev/glyssen/blob/6ab7ae793f96a2cca0e5ff587513baae95ce1878/Glyssen/UsxParser.cs#L237)):*

I was asking because if you had something like

/v 1 Some text /p -- /v 2 More text

then this code would not cover that case, I think. But that is probably an unlikely scenario.
</blockquote></details>

It depends on what you mean by "cover". In that case, the code should do nothing. I agree that it would be unusual, but we definitely wouldn't want to delete either verse in this case. Not exactly sure what the quote parser or ISP would do with data like that, but if there was a problem, we'd probably want to fix it downstream.

tombogle:
___
*[Glyssen/UsxParser.cs, line 253 at r2](https://reviewable.io/reviews/sillsdev/glyssen/490#-Lcar9V-78wjktaqAAfu:-LcawBjH5niXQ0A-qD4-:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/e613ff00de39abf63ed7bde125c2c208e5ef3e57/Glyssen/UsxParser.cs#L253)):*

> ```
> unweildy
> ```
unwieldy (if you're making changes)
</blockquote></details>

Done.

tombogle:
___
*[Glyssen/UsxParser.cs, line 254 at r2](https://reviewable.io/reviews/sillsdev/glyssen/490#-LcaqWN06ttKC3Now3XK:-LcawBPnBjVraJgO_OuU:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/e613ff00de39abf63ed7bde125c2c208e5ef3e57/Glyssen/UsxParser.cs#L254)):*

> ```
> somethign
> ```
something (if you're making changes)
</blockquote></details>

Done.

tombogle:
___
*[Glyssen/UsxParser.cs, line 235 at r2](https://reviewable.io/reviews/sillsdev/glyssen/490#-LcaqLAzB1j7xoaerbLk:-Lcaul_B3bGsMDXY7SlL:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/e613ff00de39abf63ed7bde125c2c208e5ef3e57/Glyssen/UsxParser.cs#L235)):*

> ```
> empt
> ```
empty (if you're making changes)
</blockquote></details>

Done.

PR #492, opened 2019-04-18
PG-1188: Fixed logic to handle breaking multi-block quote chains when block(s)…

tombogle:
… in matchup are assigned to narrator.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/492)


tombogle:
I don't think a data migration is needed this time because the problem actually caused Glyssen to crash before bad data could get saved.

PR #494, opened 2019-04-24
PG-1192: Fixed problem: Audio was being appended twice for the default project name.

tombogle:
This also includes a data migration to clean up the extra Sample project.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/494)

### DIFF
--- a/DistFiles/localization/Glyssen.es.tmx
+++ b/DistFiles/localization/Glyssen.es.tmx
@@ -7415,12 +7415,66 @@
         <seg>Grabando... (disculpe la tardanza)</seg>
       </tuv>
     </tu>
+    <tu tuid="DataMigration.AudioAudioProblemPreambleMultiple">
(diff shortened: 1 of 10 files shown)


tombogle:
Good catches, mostly. Actually, you're second catch wasn't quite accurate. But my code was definitely wrong. I failed to delete the following code, because in the case of the unsafe ones, I don't actually attempt to clean them up at all. I just got lucky in that DialogResult.OK never equals DialogResult.Yes :-). As for the weird thing at the end with retVal, if this method returns false, the settings don't get saved. I figured it could just skip setting the version number entirely, but I didn't take the time earlier to confirm that. I have done so now, and changed it.

tombogle:
___
*[Glyssen/DataMigrator.cs, line 289 at r2](https://reviewable.io/reviews/sillsdev/glyssen/494#-LdGRU6S8OBR7UWtmhjR:-LdHJ41A37mqXpEIlmui:b-cdu6cl) ([raw file](https://github.com/sillsdev/glyssen/blob/bdc6617cf0678a53ce01ad0d1230f54ff45cbe3f/Glyssen/DataMigrator.cs#L289)):*

> ```
> safeReplacements
> ```
Should be unsafeReplacements
</blockquote></details>

Should be ripped out entirely :-)

PR #495, opened 2019-04-25
PG-1190: Prevented loss of user decisions when upgrading a Paratext project…

tombogle:
…to a version of Glyssen that has a new parser.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/495)

### DIFF
--- a/Glyssen/Project.cs
+++ b/Glyssen/Project.cs
@@ -1177,6 +1177,8 @@ namespace Glyssen
 				}
 				if (foundDataChange)
 					upgradedProject.m_projectMetadata.Revision++; // See note on GlyssenDblTextMetadata.RevisionOrChangesetId
+
+				upgradedProject.ProjectStateChanged?.Invoke(upgradedProject, new ProjectStateChangedEventArgs());
 			};
 
 			UpgradeProject(existingProject, upgradedProject, () =>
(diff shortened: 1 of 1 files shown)


tombogle:
___
*[Glyssen/Project.cs, line 1164 at r1](https://reviewable.io/reviews/sillsdev/glyssen/495#-LdK3HZl-6CMLMSO9QkM:-LdK6DP70ngVSyV8YOkx:b-ker17i) ([raw file](https://github.com/sillsdev/glyssen/blob/8da24ece901ec34eaa56deb0530df912a98c28d5/Glyssen/Project.cs#L1164)):*

Should we be unsubscribing this somewhere?
</blockquote></details>

I haven't been able to test a scenario where it comes back through the quote parser with the same project, but it seems like it could be possible if I didn't have access to change the Paratext project and therefore overrode the quote settings directly in Glyssen. I'll put it in a method and unsubscribe

tombogle:
___
*[Glyssen/Project.cs, line 1170 at r1](https://reviewable.io/reviews/sillsdev/glyssen/495#-LdK3c_d-Xe45yrNB5Je:-LdKDjuJ63T-eMzzF4fd:b-ce6r20) ([raw file](https://github.com/sillsdev/glyssen/blob/b1fd8101be6344de45d6006d71ddc325902ad9c2/Glyssen/Project.cs#L1170)):*

> ```
> book.ParatextChecksum = scrTextWrapper.GetBookChecksum(book.BookNumber);
> ```
This is either not needed or being set during UpgradeProject?
</blockquote></details>

No, as I was looking at the previous change, I noticed that it got clobbered and put it back in. I'm guessing I just accidentally hit Ctrl-X and axed it without noticing. My confidence in my ability to make a safe change is shaken...

PR #496, opened 2019-04-25
PG-1194: Ensured that project-specific character-verse and character-…

tombogle:
…detail info are included from source project when applying user decisions.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/496)

### DIFF
--- a/Glyssen/Character/ProjectCharacterVerseData.cs
+++ b/Glyssen/Character/ProjectCharacterVerseData.cs
@@ -65,5 +65,11 @@ namespace Glyssen.Character
 			RemoveAll(GetCharacters(bookNum, chapterNumber, initialStartVerseNumber, initialEndVerseNumber)
 				.Where(c => c.Character == characterId && c.Delivery == delivery), new BcvCharacterDeliveryEqualityComparer());
 		}
+
+		public void UnionWith(ProjectCharacterVerseData otherProjectData)
+		{
+			foreach (var cv in otherProjectData.GetAllQuoteInfo())
+				AddCharacterVerse(cv);
+		}
 	}
 }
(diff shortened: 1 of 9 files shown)


tombogle:
I put that in back when I created it as a WIP PR. I've removed it now.

tombogle:
___
*[GlyssenTests/ProjectTests.cs, line 255 at r1](https://reviewable.io/reviews/sillsdev/glyssen/496#-Ldf8bG860xQFVo1Z8pT:-Ldijr5DEKPbE_r7ires:b-nieei) ([raw file](https://github.com/sillsdev/glyssen/blob/9e7dfe3028c4e631ae6805d92521af0a8d625494/GlyssenTests/ProjectTests.cs#L255)):*

> ```
> afterQuoteParserCompletes
> ```
I think this would make more sense to me if it was named something like `resultQuoteSystem` or `resolvedQuoteSystem`
</blockquote></details>

Done. (I don't actually remember adding this test and just barely see how it relates to this issue.)

tombogle:
___
*[GlyssenTests/ProjectTests.cs, line 262 at r1](https://reviewable.io/reviews/sillsdev/glyssen/496#-Ldf9ZZN2FAsMqBjTStQ:-Ldikk6a8h1UO1naAaSH:b-6dbpu0) ([raw file](https://github.com/sillsdev/glyssen/blob/9e7dfe3028c4e631ae6805d92521af0a8d625494/GlyssenTests/ProjectTests.cs#L262)):*

> ```
> project.QuoteSystem = QuoteSystem.Default;
> ```
I'm a bit confused about this test. The title claims we are either changing the quote system and/or checking that it was changed.
But the test neither proves we are changing it nor asserts it was one thing and became another.
I probably wrote the confusion in the first place, but maybe some comments could help?
</blockquote></details>

Indirectly it did because the quote parse wold only happen if the quote system actually changed. But I've made it more explicit and improved the name of the test.

PR #498, opened 2019-04-29
PG-1195: Fixed logic in GetIndexOfFirstBlockForVerse

tombogle:
---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/498)

### DIFF
--- a/Glyssen/BookScript.cs
+++ b/Glyssen/BookScript.cs
@@ -324,11 +324,13 @@ namespace Glyssen
 				if (block.InitialStartVerseNumber < verse && block.InitialEndVerseNumber < verse)
 					continue;
 				iFirstBlockToExamine = index;
-				if (block.InitialStartVerseNumber > verse || block.InitialEndVerseNumber > verse ||
-					(iFirstBlockToExamine > 0 && !(block.BlockElements.First() is Verse) && m_blocks[iFirstBlockToExamine - 1].LastVerseNum == verse))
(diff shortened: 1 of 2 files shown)


andrew-polk:
"WIP" in PR description

tombogle:
This is a condition that my fix for PG-1196 actually broke (because there were no tests covering it). Section heads have the reference of the preceding verse (which is actually logically wrong, but that's the way our USX parser assigns them. (It's easier than getting them right, and it doesn't seem to cause us any actual consternation.) The logic error I had introduced, however, was causing the index to back up and land on the section head. I considered actually using the IsScripture property (which checks the character ID), but it's not terribly efficient and since the logic doesn't really require it (because the section head is actually associated with the previous verse), I decided against it.

tombogle:
___
*[Glyssen/BookScript.cs, line 329 at r1](https://reviewable.io/reviews/sillsdev/glyssen/498#-LdfBwLV6olnzOjw3_in:-LdifZ47DmT04IzXXQsF:bqg960h) ([raw file](https://github.com/sillsdev/glyssen/blob/1ca7b7220490822f9f6b00ee9e8b45b388c70ebe/Glyssen/BookScript.cs#L329)):*

> ```
> block.InitialEndVerseNumber > 0
> ```
Always true if initial end verse number is >= verse, right? Just have it there for clarity?
</blockquote></details>

True. Although technically, we don't check to make sure verse is non-negative. :-)

PR #501, opened 2019-05-09
PG-1203: Distinguished between \qa and other "normal" poetry styles in IsFollowOnParagraphStyle

tombogle:
This prevents combining the acrostic headings in Psalm 119 with adjoining poetry paragraphs

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/501)

### DIFF
--- a/Glyssen/App.config
+++ b/Glyssen/App.config
@@ -70,7 +70,7 @@
     <applicationSettings>
         <Glyssen.Properties.Settings>
             <setting name="ParserVersion" serializeAs="String">
-                <value>45</value>
+                <value>46</value>
             </setting>
             <setting name="DataFormatVersion" serializeAs="String">
                 <value>4</value>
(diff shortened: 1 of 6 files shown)


tombogle:
___
*[Glyssen/Block.cs, line 65 at r1](https://reviewable.io/reviews/sillsdev/glyssen/501#-LeTGRxd9nRwYiynsYVz:-LeTPQy22rli7GPZoWGw:b9gbyhj) ([raw file](https://github.com/sillsdev/glyssen/blob/0a0b02f3b61d885f9ca244533ee2e1c5753df730/Glyssen/Block.cs#L65)):*

If the idea is to have all q?? markers besides qa, I wonder if there is a way to write the regex that way.
In other words, blacklist qa rather than try to whitelist the rest. Feels more future-proof.
If it isn't worth the complication, we could at least put a comment here stating the goal.
</blockquote></details>

That was my original plan and I started down that road, but when I looked into the future, I realized I had no idea what behavior we were most likely to want with the next marker they added. I think it is a little safer to only join the things we know about than to blindly join all future things. At least as it is, it's reasonably possible to look at the USFM documentation and know exactly which markers we had in mind when the code was written.

tombogle:
*[a discussion](https://reviewable.io/reviews/sillsdev/glyssen/501#-LeTYg-t543LTETGcxAk:-LeTYg-u09h_YAJai4B9:bljzc47) (no related file):*
I guess there really needs to be a new parser version since this affects the way the quote parser works.

---

PR #504, opened 2019-05-22
PG-1205: Prevent flicker when changing filters but staying on same rainbow matchup

tombogle:
In the process of looking at PG-1204, I noticed a flicker when changing filters in rainbow mode. This fix should prevent that when the existing matchup also matches the new filter.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/504)


tombogle:
Oops, looks like I broken something...

PR #507, opened 2019-05-29
PG-1210: Added delivery and paragraphStart attributes to glyssenscript

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/507)

### DIFF
--- a/Glyssen/ProjectExporter.cs
+++ b/Glyssen/ProjectExporter.cs
@@ -866,6 +866,8 @@ namespace Glyssen
 			if (includeDelivery)
 				exportData.Delivery = block.Delivery;
 
+			exportData.IsParagraphStart = block.IsParagraphStart;
+
 			if (!getBlockElements)
 			{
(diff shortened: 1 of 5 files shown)


tombogle:
___
*[GlyssenShared/Script/GlyssenScript.cs, line 20 at r1](https://reviewable.io/reviews/sillsdev/glyssen/507#-Lg3YTeN5Iqb1TYQ-IVz:-Lg3mU2d1g4jvyD9LzRf:bjaqkvp) ([raw file](https://github.com/sillsdev/glyssen/blob/2f239804100bf1fffa354ef66b80d27d9bfcc71b/GlyssenShared/Script/GlyssenScript.cs#L20)):*

I'm pretty sure we want 2.1 here. If we go to 3, I think HearThis won't open them. This isn't a breaking change, right?
</blockquote></details>

Thanks. I was trying to figure out where this was used or what its significance was. No, it's not a breaking change.

PR #515, opened 2019-07-11
PG-794: Adjusted reference text matching logic to handle some common …

tombogle:
…mismatch cases and ensure that no reference blocks are dropped when aligning.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/515)

### DIFF
--- a/Glyssen/Block.cs
+++ b/Glyssen/Block.cs
@@ -253,20 +253,30 @@ namespace Glyssen
 		[DefaultValue(-1)]
 		public int SplitId { get; set; }
 
+		/// <summary>
+		/// Don't use in production code. It is intended ONLY for use by the XML serializer!
+		/// To ensure internal consistency, use SetMatchedReferenceBlock, ClearReferenceText,
+		/// SetUnmatchedReferenceBlocks, AppendUnmatchedReferenceBlock, and
+		/// InsertUnmatchedReferenceBlocks methods.
+		/// </summary>
 		[XmlAttribute("matchesReferenceText")]
 		[DefaultValue(false)]
-		public bool MatchesReferenceText
+		public bool MatchesReferenceText_DoNotUse
 		{
 			// m_matchesReferenceText should imply exactly one reference block (and I *so* wish I had modeled it
 			// that way), but if a previous program bug or deserialization issue should put a block into a bad
 			// state, there are lots of places where it could crash, so if there isn't exactly 1, return false.
-			get => m_matchesReferenceText && ReferenceBlocks.Count ==  1;
-			set
-			{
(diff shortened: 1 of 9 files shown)


tombogle:
___
*[Glyssen/ReferenceText.cs, line 422 at r1](https://reviewable.io/reviews/sillsdev/glyssen/515#-LjY3dSy8OtbvTyru9Fd:-LjaDuVyCZslxK5A-l3K:b4s72d0) ([raw file](https://github.com/sillsdev/glyssen/blob/73a0a5d50f05439de3a6dc77b6e09327dfcf52ff/Glyssen/ReferenceText.cs#L422)):*

I'm sure you've evaluated this how likely it is to get here if we don't want this result, but it seems like a pretty bold assumption on the surface.

Any comments you could add to help clarify what some of these if conditions actually mean in English would be great. I know what you're trying to do and I still don't understand it.
</blockquote></details>

I thought about adding more conditions to better ensure that we really do have the stated condition. But when I ran the tests, I got one failure and determined that the new result was slightly better than the previously expected result, even though the exact condition was not the one I had written the code for. So maybe I just need to make the comment say that it's probably still the right thing to do in other similar circumstances that might happen to match the conditional logic. It'd be nice to know if there are any conceivable circumstances where this would be a bad thing to do, but I'm apparently not creative enough to come up with one, nor to prove conclusively that there aren't any.

tombogle:
___
*[Glyssen/ReferenceText.cs, line 488 at r1](https://reviewable.io/reviews/sillsdev/glyssen/515#-LjY28qM85tm24PVSGLF:-LjaEuQw8pl_sZvKLcFN:b-oo8ajp) ([raw file](https://github.com/sillsdev/glyssen/blob/73a0a5d50f05439de3a6dc77b6e09327dfcf52ff/Glyssen/ReferenceText.cs#L488)):*

> ```
> setting MatchesReferenceText to flase CLEARS the existing list!
> ```
Seems wrong, doesn't it? Should we make it a method instead of a property and make the name more clear?
Or should there be a different method you call which sets the property to false and clears the list?
</blockquote></details>

Yes, it does. Unfortunately, both MatchesReferenceText and ReferenceBlocks are part of the XML model, so this code is a partial attempt to try to enforce some level of consistency. Moving it into a separate method would increase the likelihood that the two things could get out of sync. And maybe that would be okay, since it's already not 100% guaranteed. I commented out the code in the MatchesReferenceText setter that clears the list, and only one test fails. So that, at least, would be easy to fix. What I'd really like to do is just define the getter as ReferenceBlocks.Count ==  1. But I'm much less confident that doing that would be safe. (I got quite a few test failures when I tried it.)

PR #517, opened 2019-07-12
PG-1133: Deal with indirect speech misalignments

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/517)


andrew-polk:
There's something wrong with this PR. It includes changes from two other commits.

tombogle:
I cherry-picked an emergency-fix commit, but then added comments around that line, resulting in a conflict, even though there was no actual change to that line. (Never had to do this before, but I had to do a commit with no changes.)

PR #521, opened 2019-08-01
PG-1235: Prevented crash resulting from bad multi-block block chains

tombogle:
when blocks that are part of a runaway open quote are assigned to narrator.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/521)


tombogle:
No cleanup needed because in this case Glyssen would crash (and not save).

PR #523, opened 2019-08-07
Added trace listener to channel trace info from ParatextData into our log file.

tombogle:
I implemented this in hopes of getting some useful info to track down ICU load error reported in PG-1234. However, the trace info from ICU do not make it out, so all we get is the trace inf from ParatextData. Still that seems like a useful improvement.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/523)

### DIFF
--- a/Glyssen/Glyssen.csproj
+++ b/Glyssen/Glyssen.csproj
@@ -380,6 +380,7 @@
       <SubType>Component</SubType>
     </Compile>
     <Compile Include="Utilities\L10N.cs" />
+    <Compile Include="Utilities\LogFileTraceListener.cs" />
     <Compile Include="Utilities\MessageBoxStrings.cs" />
     <Compile Include="Utilities\UndoActionSequence.cs" />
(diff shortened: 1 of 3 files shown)


tombogle:
___
*[Glyssen/Utilities/LogFileTraceListener.cs, line 17 at r1](https://reviewable.io/reviews/sillsdev/glyssen/523#-Llhxl3P10_kcXH3P_Yx:-LmAyE6358rlcNBt-LRz:b-hqkui8) ([raw file](https://github.com/sillsdev/glyssen/blob/2ecffa46d35aa327372c02dc8b82cbee752676a9/Glyssen/Utilities/LogFileTraceListener.cs#L17)):*

WriteMinorEvent?
Otherwise, LGTM unless this is going to fill up the log with useless goo.
</blockquote></details>

From what I can tell (and from what Tim says), I think there are very few things that are likely to get written out via Trace from ParatextData. Since the one thing I know of seems like it could be fairly useful in diagnostics, I think I'm inclined to keep this as WriteEvent.

PR #524, opened 2019-08-15
PG-1233 Added rare/potential speakers that sometimes occur

tombogle:
in languages that turn thoughts, reports, etc. into quotes. This also includes a code change to handle the new **Rare** quote type.
Note that the branch was named for another issue, but then morphed into this one.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/524)


tombogle:
I need to figure out what issue this is really for. PG-1242 seems to be something else and is not even in progress.

tombogle:
I figured out that this is indeed for PG-1242. See this comment: https://jira.sil.org/browse/PG-1242?focusedCommentId=214088&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-214088

tombogle:
Branch renamed

PR #526, opened 2019-08-15
PG-1446: Fixed several bugs which combined to prevent proper parsing of verse number…

tombogle:
…in a reference text being SplitBlock

Also improved implementation and clarity in logic designed to prevent a screwy InitialEndVerseNumber when parsing a verse number in a reference text.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/526)

### DIFF
--- a/Glyssen/Block.cs
+++ b/Glyssen/Block.cs
@@ -355,7 +355,8 @@ namespace Glyssen
 		{
 			if (referenceBlocksToJoin == null)
 				referenceBlocksToJoin = ReferenceBlocks;
-			var refBlock = new Block(StyleTag, ChapterNumber, InitialStartVerseNumber, InitialEndVerseNumber);
+			var baseBlock = (referenceBlocksToJoin?.FirstOrDefault() ?? this);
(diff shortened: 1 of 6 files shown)


tombogle:
___
*[GlyssenTests/BlockMatchupTests.cs, line 898 at r1](https://reviewable.io/reviews/sillsdev/glyssen/526#-LmQ1zsV3McQPwDGWMPZ:-LmQSOXSBenhsLXHUCN5:braxfwv) ([raw file](https://github.com/sillsdev/glyssen/blob/816afecd9b4ebe990dd608f0a05ac6c830a37a88/GlyssenTests/BlockMatchupTests.cs#L898)):*

I'm not following how we should end up with a 1-2 verse bridge.
But it if makes sense to you, :thumbsup: 
</blockquote></details>

The logic is entirely "fuzzy", but the idea is that since the corresponding vernacular block is a verse bridge from 1-3, that the reference text is "probably" for that same range. But since it turns out that the reference text has a 3-4 verse bridge in the middle of it, logically the preceding text can't be for verse 1-3, so they must be for verses 1-2. If that even comes close to making sense, I can add it as a comment.

PR #528, opened 2019-08-30
PG-1257: Fixed crash when adding project-specific control file entries

tombogle:
This addresses both the underlying cause of the crash reported in PG-1257 as well as adding the missing entry (for MAT 27:39) to the factory control file.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/528)

### DIFF
--- a/Glyssen/Block.cs
+++ b/Glyssen/Block.cs
@@ -201,6 +201,8 @@ namespace Glyssen
 			/// If the Verse number represents a verse bridge, this will be the ending number in the bridge; otherwise 0.
 			/// </summary>
 			public int LastVerseOfBridge { get; private set; }
+
+			public IEnumerable<int> AllVerseNumbers => this.GetAllVerseNumbers();
 		}
 
 		public int LastVerseNum => LastVerse.EndVerse;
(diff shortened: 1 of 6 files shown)


tombogle:
___
*[Glyssen/Resources/CharacterVerse.txt, line 13869 at r1](https://reviewable.io/reviews/sillsdev/glyssen/528#-LnZfi9r04G_2PopVzXv:-LnsEWxyAPL2-_0iWTqH:bkcto2u) ([raw file](https://github.com/sillsdev/glyssen/blob/a98c0d02f30ea0c710988ece0c8fe08af4d79f70/Glyssen/Resources/CharacterVerse.txt#L13869)):*

Hm. This was in that other PR, too.
Version bump?
</blockquote></details>

It was only supposed to be in this one, not the other. I didn't bother with a version bump because it seems no existing project could have added this since it caused a crash. (Trying to avoid unnecessary version bumps conflicting with the merges from the other big changes.)

tombogle:
___
*[GlyssenTests/Dialogs/AssignCharacterViewModelTests.cs, line 363 at r1](https://reviewable.io/reviews/sillsdev/glyssen/528#-LnZiAIm2hHrXVvWxjGh:-Lnw5haDA-uBck57772M:b-66lnwr) ([raw file](https://github.com/sillsdev/glyssen/blob/a98c0d02f30ea0c710988ece0c8fe08af4d79f70/GlyssenTests/Dialogs/AssignCharacterViewModelTests.cs#L363)):*

Should have read, "comment on what is in the control file."
</blockquote></details>

Unlike the tests that operate against the real live control files, this fixture uses the TestCharacterVerse file, so presumably no change will ever be made to that resource without first ensuring no negative impact on any existing test. So hitherto we haven't tried to document the relevant contents of the file for each test, though in some cases we have done the up-front verification to prove that it has what we want. I can add lines to do that.

tombogle:
___
*[Glyssen/Dialogs/AssignCharacterViewModel.cs, line 148 at r1](https://reviewable.io/reviews/sillsdev/glyssen/528#-LnZiUzO8h_u90tIFp9K:-LnwJwmy76pSd7AUXjbR:bcp1lkf) ([raw file](https://github.com/sillsdev/glyssen/blob/a98c0d02f30ea0c710988ece0c8fe08af4d79f70/Glyssen/Dialogs/AssignCharacterViewModel.cs#L148)):*
> ```
> / REVIEW: Probably not strictly necessary, but shouldn't we really add one per verse if this bloc
> ```
Someday... I looked at it, and there's a bit of refactoring (and re-thinking) to do because GetUniqueCharacterVerseObjectsForBlock needs to be re-written (or probably split up into two different implementations for it to actually work as desired.

PR #532, opened 2019-09-10
PG-1261: Fixed logic error in IsRelevant that caused it to consider a…

tombogle:
… block relevant in a case where ProjectAnalysis.CalculateAlignmentPercentage does not.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/532)


tombogle:
___
*[GlyssenTests/Dialogs/BlockNavigatorViewModelTests.cs, line 1343 at r1](https://reviewable.io/reviews/sillsdev/glyssen/532#-LoS8Jr_8Y8X7JoO7Me4:-LoVECiE9N9AhMpn2tSF:b-x1j042) ([raw file](https://github.com/sillsdev/glyssen/blob/0f82e1cbeafa07df54963f96da3aad5753ef72b6/GlyssenTests/Dialogs/BlockNavigatorViewModelTests.cs#L1343)):*

I'll have to trust you that the user doesn't want to be taken to such a place to investigate.
If so, LGTM
</blockquote></details>

No, actually, that comment was incorrect. The user DOES want to be taken there!

PR #535, opened 2019-09-19
PG-1262: Changed UsxParser to handle odd verse bridges

tombogle:
PG-1262: It's not entirely clear whether in some past version of the standard it was considered acceptable to indicate a verse bridge formed by two verse fields with a dash in the text of the first one, but since Paratext does not flag it as an error, I've modified the USX Parser to handle it.

PG-1263: Added check to try to track down mysterious ArgumentOutOfRangeException exception.
Also did a little bit of refactoring and typo cleanup

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/535)


tombogle:
___
*[Glyssen/Dialogs/BlockNavigatorViewModel.cs, line 944 at r1](https://reviewable.io/reviews/sillsdev/glyssen/535#-Lpo_Dsa73_u7c5hMLFn:-Lq1hUB39CYgK0w_24IG:b-l11iax) ([raw file](https://github.com/sillsdev/glyssen/blob/553dde1a8431264c4f30be9c9d051d23ac6eb804/Glyssen/Dialogs/BlockNavigatorViewModel.cs#L944)):*
> ```
> throw new IndexOutOfRangeException($"Start index should never be negative. " +
> ```
The log gives us all that information already

PR #548, opened 2019-10-17
PG-1272: Interpret "wj" and "qt" character styles in usx parser

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/548)


tombogle:
___
*[Glyssen/UsxParser.cs, line 104 at r1](https://reviewable.io/reviews/sillsdev/glyssen/548#-LrPXla3B2Qlg5r8Hjgq:-LrPe1tW78qSVp_uW6pI:b-buc4o2) ([raw file](https://github.com/sillsdev/glyssen/blob/13d5ab50eb98bbf3d64cd919c241ff63c493859a/Glyssen/UsxParser.cs#L104)):*

> ```
> ["qt"] = "scripture"
> ```
I'm a little unsure about this one....
The spec says, "Old Testament quotations in the New Testament, or other quotations."
I guess the idea is it is okay because we are checking if the control file has scripture for that verse? That's reasonable. 
</blockquote></details>

Yes, that was my rationale. Of course, there's always the possibility that a verse could have both a potential Scripture quote and a quote of someone else, but it's probably not too likely (and then the translation team would have had to mark it up with \qt, which isn't a very common style anyway). Even then, there's a pretty good chance it would result in a rainbow mismatch.

tombogle:
___
*[Glyssen/UsxParser.cs, line 216 at r1](https://reviewable.io/reviews/sillsdev/glyssen/548#-LrPZH_O6EH7gppaaPFT:-LrPedLABty6SdJgMDVd:b-euzh98) ([raw file](https://github.com/sillsdev/glyssen/blob/13d5ab50eb98bbf3d64cd919c241ff63c493859a/Glyssen/UsxParser.cs#L216)):*

The quote parser doesn't overwrite these, I guess?
</blockquote></details>

Working on that now...

tombogle:
___
*[GlyssenTests/TestProject.cs, line 144 at r1](https://reviewable.io/reviews/sillsdev/glyssen/548#-LrPXBOqEUyCcrVCoHhQ:-LrPeiqKEOWpk0fLymkw:b-helncx) ([raw file](https://github.com/sillsdev/glyssen/blob/13d5ab50eb98bbf3d64cd919c241ff63c493859a/GlyssenTests/TestProject.cs#L144)):*

Any reason to leave this here?
</blockquote></details>

The test that calls it is also commented out. It is something we're probably not terribly likely to do (and when we do, we may not happen upon the commented out test...), but I figured that until we decide for sure to rip it out, this code should probably also remain as commented out code.

tombogle:
___
*[Glyssen/UsxParser.cs, line 216 at r1](https://reviewable.io/reviews/sillsdev/glyssen/548#-LrPZH_O6EH7gppaaPFT:-LrQXg4s1kJBjHr06Jzo:bp9lalq) ([raw file](https://github.com/sillsdev/glyssen/blob/13d5ab50eb98bbf3d64cd919c241ff63c493859a/Glyssen/UsxParser.cs#L216)):*

Working on that now...
</blockquote></details>

Okay, I think everything is done and working now. I did add one REVIEW comment to one of the QuoteParser tests, but I'm not quite sure if I can figure out how to know if I've opened up a loophole that could lead to something undesirable.

tombogle:
This is going to be dependent on https://github.com/sillsdev/Glyssen/pull/549

tombogle:
___
*[Glyssen/UsxParser.cs, line 216 at r1](https://reviewable.io/reviews/sillsdev/glyssen/548#-LrPZH_O6EH7gppaaPFT:-LssP2qYA9eT8V_8ltIh:b-8orf0q) ([raw file](https://github.com/sillsdev/glyssen/blob/13d5ab50eb98bbf3d64cd919c241ff63c493859a/Glyssen/UsxParser.cs#L216)):*

Seems okay to me. But I might not understand the implication you raise regarding the project-specific file.
Generally, I would say we should avoid complexity due to delivery at all cost since they aren't even using it yet.
</blockquote></details>

Yeah, my hope with this change was to reduce complexity for the user by not forcing them to make an impossible decision about which delivery to use when neither one made sense and certainly not to force it to use the wrong one. Since the quote parser currently never tries to add a project-specific CV entry, my concern is/was that the data migrator could remove something set by the quote parser because it can't find a matching delivery in either the project-specific or the control-file based CV data.

PR #550, opened 2019-10-18
PG-1276+ Prevent conflict between parser and ISP; Use undefined delivery when block has verses with conflicting deliveries

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/550)

### DIFF
--- a/ControlDataIntegrityTests/CharacterVerseDataTests.cs
+++ b/ControlDataIntegrityTests/CharacterVerseDataTests.cs
@@ -4,6 +4,8 @@ using System.Linq;
 using System.Text.RegularExpressions;
 using Glyssen.Character;
 using Glyssen.Properties;
+using Glyssen.Shared;
+using Glyssen.Utilities;
 using NUnit.Framework;
 using SIL.Scripture;
 
@@ -99,17 +101,27 @@ namespace ControlDataIntegrityTests
 		}
 
 		[Test]
-		public void DataIntegrity_NoDuplicateWhereOnlyDifferenceIsNormalVsNonnormalDelivery()
+		public void DataIntegrity_NoDuplicateWhereOnlyDifferenceIsNormalVsNonNormalDelivery()
 		{
-			// PG-152: Currently, the program does not handle duplicates where the
-			// only difference is between normal (blank) delivery and a specified delivery
-			ISet<CharacterVerse> uniqueCharacterVerses = new HashSet<CharacterVerse>(new BcvCharacterAndTypeEqualityComparer());
+			// PG-152/PG-1272: Glyssen now handles duplicates where the only difference is
+			// between normal (blank) delivery and a specified delivery, but if the same
+			// character speaks twice in the same verse with two different deliveries, it's
+			// still genrally best to make them both explicit. We allow for the exception
+			// of an Alternate (where the main speaker is allowed to speak the quotation
+			// rather than having it spoken by the character being quoted), because in this
+			// case the delivery is as much an informational note to the scripter as it is
+			// a delivery for the recording team to attend to.
+			IEqualityComparer <CharacterVerse> comparer = new BcvCharacterEqualityComparer();
+			ISet<CharacterVerse> uniqueCharacterVerses = new HashSet<CharacterVerse>(comparer);
 			IList<CharacterVerse> duplicateCharacterVerses = new List<CharacterVerse>();
 			foreach (CharacterVerse cv in ControlCharacterVerseData.Singleton.GetAllQuoteInfo()
 				.OrderBy(cv => cv.BcvRef).ThenBy(cv => string.IsNullOrEmpty(cv.Delivery)))
 			{
 				if (!uniqueCharacterVerses.Add(cv) && string.IsNullOrEmpty(cv.Delivery))
-					duplicateCharacterVerses.Add(cv);
+				{
+					if (uniqueCharacterVerses.Single(c => comparer.Equals(c, cv)).QuoteType != QuoteType.Alternate)
+						duplicateCharacterVerses.Add(cv);
+				}
 			}
 
 			Assert.False(duplicateCharacterVerses.Any(),
@@ -147,6 +159,12 @@ namespace ControlDataIntegrityTests
 				if (!charactersHavingDetail.Contains(cv.Character))
 				{
 					if (CharacterVerseData.IsCharacterStandard(cv.Character) || cv.Character == CharacterVerseData.kNeedsReview)
+						continue;
+					// For interruptions, we actually never pay any attention to the character ID. They are always narrator.
+					// But in some verses we have both a normal narrator Quotation and an interruption, so to avoid a problem
+					// in other data integrity checks, we just use the dummy character id "interruption-XXX" (where XXX is the
+					// 3-letter book code.
+					if (cv.QuoteType == QuoteType.Interruption || cv.Character.StartsWith("interruption-"))
 						continue;
 
 					var characters = cv.Character.Split('/');
@@ -228,10 +246,10 @@ namespace ControlDataIntegrityTests
 		public void DataIntegrity_ImplicitCharacterIsExclusiveUnlessItIsNeedsReviewOrTheOtherEntryIsAQuotationInDeuteronomy()
 		{
 			foreach (var cv in ControlCharacterVerseData.Singleton.GetAllQuoteInfo()
-				.Where(i => i.QuoteType == QuoteType.Implicit && i.Character != CharacterVerseData.kNeedsReview))
+				.Where(i => i.IsImplicit && i.Character != CharacterVerseData.kNeedsReview))
 			{
 				var otherEntries = ControlCharacterVerseData.Singleton.GetCharacters(BCVRef.BookToNumber(cv.BookCode),
-					cv.Chapter, cv.Verse, versification: ScrVers.English)
+					cv.Chapter, new SingleVerse(cv.Verse), ScrVers.English)
 					.Where(c => c != cv && c.QuoteType != QuoteType.Hypothetical &&
 					(c.QuoteType != QuoteType.Quotation || c.Character != cv.Character)).ToList();
 				if (otherEntries.Any())
@@ -263,16 +281,18 @@ namespace ControlDataIntegrityTests
 				QuoteType.Potential,
 				QuoteType.Quotation,
 				QuoteType.Hypothetical,
-				QuoteType.Implicit
+				QuoteType.Implicit,
+				QuoteType.ImplicitWithPotentialSelfQuote
 			};
+			var numberOfAcceptableQuoteTypesInOT = acceptablePrimaryQuoteTypes.Count;
 			foreach (var alternate in ControlCharacterVerseData.Singleton.GetAllQuoteInfo()
 				.Where(i => i.QuoteType == QuoteType.Alternate))
 			{
-				if (alternate.BookCode == "MAT" && acceptablePrimaryQuoteTypes.Count == 5)
+				if (alternate.BookCode == "MAT" && acceptablePrimaryQuoteTypes.Count == numberOfAcceptableQuoteTypesInOT)
 					acceptablePrimaryQuoteTypes.Remove(QuoteType.Hypothetical);
 				var otherEntries = ControlCharacterVerseData.Singleton.GetCharacters(BCVRef.BookToNumber(alternate.BookCode),
-						alternate.Chapter, alternate.Verse, versification: ScrVers.English)
-					.Where(c => acceptablePrimaryQuoteTypes.Contains(c.QuoteType));
+						alternate.Chapter, new SingleVerse(alternate.Verse), ScrVers.English)
+					.Where(c => acceptablePrimaryQuoteTypes.Contains(c.QuoteType)).ToList();
 				Assert.IsFalse(otherEntries.Any(c => c.Character == alternate.Character && c.Delivery == alternate.Delivery),
 					$"Alternate used for a {alternate.Character} in {alternate.BookCode} {alternate.Chapter}:{alternate.Verse}, " +
 					"but that character also has another quote type in that verse!");
@@ -309,7 +329,7 @@ namespace ControlDataIntegrityTests
 				.Where(i => i.QuoteType == QuoteType.Rare))
 			{
 				var nonRare = ControlCharacterVerseData.Singleton.GetCharacters(BCVRef.BookToNumber(rare.BookCode),
-					rare.Chapter, rare.Verse, versification: ScrVers.English, includeAlternatesAndRareQuotes:true)
+					rare.Chapter, new SingleVerse(rare.Verse), ScrVers.English, true)
 					.Where(c => c.QuoteType != QuoteType.Rare || c.Character == CharacterVerseData.kNeedsReview).ToList();
 				try
 				{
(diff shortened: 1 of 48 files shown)


tombogle:
___
*[Glyssen/ProjectDataMigrator.cs, line 264 at r1](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrtUaBq2ksQVj5GpSKT:-LruXG6E0yStvM2id4VQ:b-13lu2s) ([raw file](https://github.com/sillsdev/glyssen/blob/1aeabacf89af1512fde5050d89c48674b536381b/Glyssen/ProjectDataMigrator.cs#L264)):*
> ```
> (IEnumerable<CharacterSpeakingMode>)cvInfo.
> ```
No, I think I was being dense. Probably a hold-over from some other edit.

tombogle:
___
*[Glyssen/Character/CharacterVerse.cs, line 266 at r1](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrtTOuvDCS2mcfWxGbf:-LruXmxw9PlHlSCV7ReO:b-vljxbb) ([raw file](https://github.com/sillsdev/glyssen/blob/1aeabacf89af1512fde5050d89c48674b536381b/Glyssen/Character/CharacterVerse.cs#L266)):*

Why not quote type? 

Related, it might be helpful to add class comments for CharacterVerse and CharacterSpeakingMode.
</blockquote></details>

I'm checking to see whether this ever even gets used. I had hoped I could use it in place of having to define a CharacterDeliveryAliasEqualityComparer, but I still had to create that comparer. I really didn't want to include QuoteType in this class (because in most cases it ends up being meaningless and multiple quote types may occasionally get coalesced - though that is rare in practice), but there were places where the caller really wanted to have access to it.

tombogle:
___
*[Glyssen/Character/CharacterVerse.cs, line 266 at r1](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrtTOuvDCS2mcfWxGbf:-Lrub_mV-JYfKkhy-GWJ:b-j4t2m6) ([raw file](https://github.com/sillsdev/glyssen/blob/1aeabacf89af1512fde5050d89c48674b536381b/Glyssen/Character/CharacterVerse.cs#L266)):*

I'm checking to see whether this ever even gets used. I had hoped I could use it in place of having to define a CharacterDeliveryAliasEqualityComparer, but I still had to create that comparer. I really didn't want to include QuoteType in this class (because in most cases it ends up being meaningless and multiple quote types may occasionally get coalesced - though that is rare in practice), but there were places where the caller really wanted to have access to it.
</blockquote></details>

I have added some class comments.

tombogle:
___
*[Glyssen/Character/CharacterVerse.cs, line 352 at r1](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrtTGXk3lxeSfD4B6c0:-Lrud3Ny5uzeKvNWzMcg:b-pe8wsp) ([raw file](https://github.com/sillsdev/glyssen/blob/1aeabacf89af1512fde5050d89c48674b536381b/Glyssen/Character/CharacterVerse.cs#L352)):*

Should the fields here and in equals be the same?
I think you need the base hash as a starting point?
</blockquote></details>

I don't think it has to be the "starting point", but you're right that it should be called (rather than re-created as I was doing). From the docs: "If you override the GetHashCode method, you should also override Equals, and vice versa. If your overridden Equals method returns true when two objects are tested for equality, your overridden GetHashCode method must return the same value for the two objects."

tombogle:
___
*[GlyssenTests/ProjectDataMigratorTests.cs, line 993 at r1](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrtXZB17pJpEg6zmklI:-Lrudg7a2TYvfRIQ8se-:b-bytk3g) ([raw file](https://github.com/sillsdev/glyssen/blob/1aeabacf89af1512fde5050d89c48674b536381b/GlyssenTests/ProjectDataMigratorTests.cs#L993)):*

Pretty scary that we decided on this explicit behavior previously, even to the point of ensuring the migrator did it.
Did you look at the history of this test being added?
</blockquote></details>

I didn't look at the history. I talked to you, and you said this change sounded good. :-) I think the point of putting it in the migrator was to ensure consistency. Might be worth more study though.

tombogle:
___
*[Glyssen/Character/CharacterVerse.cs, line 266 at r1](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrtTOuvDCS2mcfWxGbf:-Lry-EFK1FLvMYOeaGte:b55hnwy) ([raw file](https://github.com/sillsdev/glyssen/blob/1aeabacf89af1512fde5050d89c48674b536381b/Glyssen/Character/CharacterVerse.cs#L266)):*

I have added some class comments.
</blockquote></details>

I have finished evaluating the equality situation and have made some needed changes to the code, the tests, the control file, and the comments. I think  you'll be satisfied.

tombogle:
___
*[GlyssenTests/ProjectDataMigratorTests.cs, line 993 at r1](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrtXZB17pJpEg6zmklI:-Lry5aPX7AkRtgGRQF7s:b-mabbgs) ([raw file](https://github.com/sillsdev/glyssen/blob/1aeabacf89af1512fde5050d89c48674b536381b/GlyssenTests/ProjectDataMigratorTests.cs#L993)):*

I didn't look at the history. I talked to you, and you said this change sounded good. :-) I think the point of putting it in the migrator was to ensure consistency. Might be worth more study though.
</blockquote></details>

I looked, and this test was added in March 2018. I was a bit surprised how recent it was. But, of course, either way, it's a fairly rare edge case -- at least it should be -- that we have a speech block covering multiple verses where the speaking character is not known to speak in all the covered verses. I'm satisfied that the significance of this test is really just about keeping consistency. However, in playing around with this specific example in Glyssen, I see there is some more work to do...

tombogle:
MigrateDeprecatedCharacterIds_FirstVerseInQuoteIsUnexpectedForCharacter_CharacterIdSetToUnknown still needs more thought. The non-rainbow mode definitely does NOT add a project entry for a block that is being assigned to a known character/delivery that is valid for one of the verses in the block, but not the first one.

tombogle:
___
*[Glyssen/Resources/CharacterVerse.txt, line 14346 at r3](https://reviewable.io/reviews/sillsdev/glyssen/550#-LryxuOJ3k-CHXotJDPP:-Lrzb3kiB5v4x-oKngBW:b-6vefp3) ([raw file](https://github.com/sillsdev/glyssen/blob/d352d5e08ac06e8e287446b425cb5a29b32f54b6/Glyssen/Resources/CharacterVerse.txt#L14346)):*

Is this supposed to be chapter 9 or 10?
</blockquote></details>

10. Probably should have an integrity test that makes sure these are in ascending order.

tombogle:
___
*[Glyssen/Resources/CharacterVerse.txt, line 14346 at r3](https://reviewable.io/reviews/sillsdev/glyssen/550#-LryxuOJ3k-CHXotJDPP:-Lrzb9Z6DjpKHklD05o3:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/d352d5e08ac06e8e287446b425cb5a29b32f54b6/Glyssen/Resources/CharacterVerse.txt#L14346)):*

10. Probably should have an integrity test that makes sure these are in ascending order.
</blockquote></details>

Done.

tombogle:
___
*[Glyssen/ProjectDataMigrator.cs, line 256 at r3](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrywXxm8UNwz05BjWbU:-Ls2PAzB08JyLseV7q_I:biefz4y) ([raw file](https://github.com/sillsdev/glyssen/blob/d352d5e08ac06e8e287446b425cb5a29b32f54b6/Glyssen/ProjectDataMigrator.cs#L256)):*

but but but...
The signature says this should be initial end verse
</blockquote></details>

So what we maybe really want is for the method to accept a list of IVerse and a flag that tells it whether to get a union or an intersection. Currently the Initial start/end pair is a range in which we get a union. If "lastVerse" is specified, then we get the intersection across that initial pair, plus each verse all the way through lastVerse.

tombogle:
___
*[Glyssen/ProjectDataMigrator.cs, line 256 at r3](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrywXxm8UNwz05BjWbU:-Ls2Q83K0P-k7QrMLE13:b-rt6p5m) ([raw file](https://github.com/sillsdev/glyssen/blob/d352d5e08ac06e8e287446b425cb5a29b32f54b6/Glyssen/ProjectDataMigrator.cs#L256)):*

So what we maybe really want is for the method to accept a list of IVerse and a flag that tells it whether to get a union or an intersection. Currently the Initial start/end pair is a range in which we get a union. If "lastVerse" is specified, then we get the intersection across that initial pair, plus each verse all the way through lastVerse.
</blockquote></details>

We probably don't need the flag. Maybe we always want an intersection. If you really want a union, then you just pass a single IVerse representing the start and end.

tombogle:
___
*[Glyssen/ProjectDataMigrator.cs, line 256 at r3](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrywXxm8UNwz05BjWbU:-Ls2QS7Y-cB1kI8DR4u6:b-78r750) ([raw file](https://github.com/sillsdev/glyssen/blob/d352d5e08ac06e8e287446b425cb5a29b32f54b6/Glyssen/ProjectDataMigrator.cs#L256)):*

We probably don't need the flag. Maybe we always want an intersection. If you really want a union, then you just pass a single IVerse representing the start and end.
</blockquote></details>

And if we made Block to have an implicit cast to IEnumerable<IVerse>, then it could make most of the calls look pretty intuitive.

tombogle:
___
*[Glyssen/ProjectDataMigrator.cs, line 256 at r3](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrywXxm8UNwz05BjWbU:-Ls2QlF_6qTW86PqgBBa:bieyi5e) ([raw file](https://github.com/sillsdev/glyssen/blob/d352d5e08ac06e8e287446b425cb5a29b32f54b6/Glyssen/ProjectDataMigrator.cs#L256)):*

And if we made Block to have an implicit cast to IEnumerable<IVerse>, then it could make most of the calls look pretty intuitive.
</blockquote></details>

IEnumerable<IVerse>

tombogle:
___
*[Glyssen/ProjectDataMigrator.cs, line 256 at r3](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrywXxm8UNwz05BjWbU:-Ls2QrJu84OqOft33gAh:b-glfnta) ([raw file](https://github.com/sillsdev/glyssen/blob/d352d5e08ac06e8e287446b425cb5a29b32f54b6/Glyssen/ProjectDataMigrator.cs#L256)):*

IEnumerable<IVerse>
</blockquote></details>

Apparently, Reviewable strips out anything wrapped in angle brackets. IEnumerable of IVerse

tombogle:
___
*[GlyssenTests/Character/CharacterVerseDataTests.cs, line 75 at r1](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrtYjjV8JAz4clnrl3l:-Ls2Rmeh8Oec-gU0QcIy:b8sauzo) ([raw file](https://github.com/sillsdev/glyssen/blob/a1d592749ec80f276af1263d032bd4c8c2c2654b/GlyssenTests/Character/CharacterVerseDataTests.cs#L75)):*

Similar to my comment in the migrator tests, it is scary that we wrote tests specifically for this scenario and now we are saying that logic is wrong. Makes me feel like we are missing something.
</blockquote></details>

Well, that is the essence of the change. Either way, it is an edge case, so it makes it a little hard to know what might be falling through the cracks. We're sealing some old cracks and maybe opening up some new ones... I did find one interesting case late yesterday that I now need to consider.

tombogle:
___
*[Glyssen/ProjectDataMigrator.cs, line 256 at r3](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrywXxm8UNwz05BjWbU:-Ls3PSxYBV8l5yLwGGeL:b70ha4e) ([raw file](https://github.com/sillsdev/glyssen/blob/d352d5e08ac06e8e287446b425cb5a29b32f54b6/Glyssen/ProjectDataMigrator.cs#L256)):*

What if you just changed the method signatures to be firstVerse and lastVerse? Or something similar.
</blockquote></details>

There are 4 things I love about the KJV: 1) It has no verse bridges; 2) it has no missing verses; 3) It uses the English versification system; and 4) The entire thing can be read using the Delivery "ancient English accent".

tombogle:
___
*[Glyssen/ProjectDataMigrator.cs, line 256 at r3](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrywXxm8UNwz05BjWbU:-Ls3Q1GG2WImdU0bbQa6:b-3pg7gp) ([raw file](https://github.com/sillsdev/glyssen/blob/d352d5e08ac06e8e287446b425cb5a29b32f54b6/Glyssen/ProjectDataMigrator.cs#L256)):*

There are 4 things I love about the KJV: 1) It has no verse bridges; 2) it has no missing verses; 3) It uses the English versification system; and 4) The entire thing can be read using the Delivery "ancient English accent".
</blockquote></details>

When I get to heaven and everyone else is hearing "Well done, good and faithful servant," I imagine God will take me over into a corner and say, "How many times did you re-write GetCharacters?!"

tombogle:
___
*[Glyssen/ProjectDataMigrator.cs, line 256 at r3](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrywXxm8UNwz05BjWbU:-LsPSJMb9nESqFspLSWm:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/d352d5e08ac06e8e287446b425cb5a29b32f54b6/Glyssen/ProjectDataMigrator.cs#L256)):*

Will you have been a better and faithfuler servant if you answer high or low?
</blockquote></details>

Done.

tombogle:
___
*[Glyssen/Resources/CharacterVerse.txt, line 14346 at r3](https://reviewable.io/reviews/sillsdev/glyssen/550#-LryxuOJ3k-CHXotJDPP:-LsPSWcM8u5XfoShNSl5:b8dqt7p) ([raw file](https://github.com/sillsdev/glyssen/blob/d352d5e08ac06e8e287446b425cb5a29b32f54b6/Glyssen/Resources/CharacterVerse.txt#L14346)):*

Done.
</blockquote></details>

And the funny thing is we *do* have such an integrity test. I saw it today. So I guess it doesn't handle this case properly.

tombogle:
___
*[Glyssen/Block.cs, line 611 at r4](https://reviewable.io/reviews/sillsdev/glyssen/550#-LsTe0-q6Gy5v1us9OGp:-LscGVeFDgMXSrwp0af0:b-26ye1g) ([raw file](https://github.com/sillsdev/glyssen/blob/d71d22b0dd4eff9cfc7e8a628991a96015b70e9f/Glyssen/Block.cs#L611)):*

For the life of me, I can't understand the syntax of this line.
</blockquote></details>

Returns an empty array. Unfortunately, I had to change AllVerses from an IEnumerable to an IReadOnlyCollection (which is really fine since every caller of it always processed them all), and so yield returns are not allowed. However, after studying the situation, I realized that there was only one place in production code where I'm ever calling this on non-Scripture blocks, and that code could be made more efficient by checking up-front and not calling this for non-Scripture blocks. So now I've changed this to throw an exception if called on a non-Scripture block.

tombogle:
___
*[Glyssen/Character/CharacterVerseData.cs, line 302 at r4](https://reviewable.io/reviews/sillsdev/glyssen/550#-LsTiykA2dmkUL-H_SzL:-LscWaeKBPbGU7JeI0eM:b-w0xh4a) ([raw file](https://github.com/sillsdev/glyssen/blob/d71d22b0dd4eff9cfc7e8a628991a96015b70e9f/Glyssen/Character/CharacterVerseData.cs#L302)):*

Should you only do this if `removed == true`?
</blockquote></details>

I think so. I could have sworn I did that actually. Maybe it was in some errant approach that I abandoned.

tombogle:
___
*[GlyssenTests/Character/CharacterVerseDataTests.cs, line 75 at r1](https://reviewable.io/reviews/sillsdev/glyssen/550#-LrtYjjV8JAz4clnrl3l:-Lscb6LL9nFUYxYrzMeh:b2nhswy) ([raw file](https://github.com/sillsdev/glyssen/blob/a1d592749ec80f276af1263d032bd4c8c2c2654b/GlyssenTests/Character/CharacterVerseDataTests.cs#L75)):*

You've done this now, I guess?
</blockquote></details>

Yes

PR #551, opened 2019-10-23
PG-1277: Ensure alignment with the reference text when a verse…

tombogle:
…is missing in the vernacular and in the reference text that verse is combined with the following verse

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/551)


tombogle:
Don't know why I keep doing that.

tombogle:
___
*[Glyssen/BlockMatchup.cs, line 284 at r1](https://reviewable.io/reviews/sillsdev/glyssen/551#-LuEId_c85Iqr1ZwJp6X:-LuJ2rwv1IZyIc90ceJq:besklx4) ([raw file](https://github.com/sillsdev/glyssen/blob/48bad6ec26c6abc40f16b1526bca1c3dd9c8df85/Glyssen/BlockMatchup.cs#L284)):*

Do we need  a test for this?
</blockquote></details>

It might not hurt to include a direct test for this case since it is a public method, but one of the existing tests that broke (namely, ApplyTo_VernacularAndReferenceTextHaveSectionHeadsInDifferentPlaces_SectionHeadReferenceTextsAreNotMatched) was fixed by adding this.

tombogle:
___
*[GlyssenTests/ReferenceTextTests.cs, line 3615 at r1](https://reviewable.io/reviews/sillsdev/glyssen/551#-LuEDkN7DrHHw0TqX5EC:-LuJ5q-O0KoBWR1zwF4I:b-xwccei) ([raw file](https://github.com/sillsdev/glyssen/blob/48bad6ec26c6abc40f16b1526bca1c3dd9c8df85/GlyssenTests/ReferenceTextTests.cs#L3615)):*

Not worth a change, but leaving the `allowSplitting: true` is just as good documentation, IMO
</blockquote></details>

I could see it either way. The disadvantages of putting in a true in the code (as opposed to using a comment) is that it is much easier for someone to come along later and remove it (since Resharper suggests it), and if the default ever changes, this test would hopefully fail, thereby perhaps alerting the programmer to the need to carefully review other places in the code that might have depended on the default of true.

PR #553, opened 2019-11-01
PG-1287 [WIP]: Prevent NullReferenceException in Select Books when PT proj has mismatched ID

tombogle:
Just preventing the crash is straightforward, but more work is needed to try to allow user to change to use the new ID.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/553)


tombogle:
Note that one of the tasks for which this PR was previously marked "WIP" has been deferred until Paratext 9.1 beta is available. (See https://jira.sil.org/browse/PG-1331)

tombogle:
___
*[Glyssen/Dialogs/ScriptureRangeSelectionDlg.cs, line 185 at r1](https://reviewable.io/reviews/sillsdev/glyssen/553#-M-phHalD1cW2vPyWMnl:-M-tw0x870ypcJedB_jH:b6ydy80) ([raw file](https://github.com/sillsdev/glyssen/blob/2e04711c5bbdbcd812a9c71708db0e3c435a3883/Glyssen/Dialogs/ScriptureRangeSelectionDlg.cs#L185)):*

I don't understand all these new early returns.
If you verify the logic and want to merge, go for it. Otherwise, I think we might need an interactive review.
Perhaps the name of the method is tripping me up? It seems that trying to get something if we need it but not getting it shouldn't have such a dramatic effect, maybe?
</blockquote></details>

I had to look back in the file history to see what this was all about because I made these changes back in November! The problem was that in the Select Books dialog, there were a bunch of places where we were calling GetParatextScrTextWrapperIfNeeded and then blindly assuming that it would have succeeded in setting m_paratextScrTextWrapper. So then the calling code would access it and throw an exception. However, the focus of this dialog is not on getting updated content from Paratext, so we don't want to bother the user if that isn't possible. (If it is possible, then we ask them about including the updated content (when appropriate). See if the changes make sense in light of that explanation. If not, we can go through them interactively. If you think this explanation (or some modified version of it) belong as a comment in the code, I could add a summary comment on GetParatextScrTextWrapperIfNeeded.

tombogle:
___
*[Glyssen/Dialogs/ScriptureRangeSelectionDlg.cs, line 185 at r1](https://reviewable.io/reviews/sillsdev/glyssen/553#-M-phHalD1cW2vPyWMnl:-M-u5c7t-HKDs38Y05rJ:b-pgmiw0) ([raw file](https://github.com/sillsdev/glyssen/blob/2e04711c5bbdbcd812a9c71708db0e3c435a3883/Glyssen/Dialogs/ScriptureRangeSelectionDlg.cs#L185)):*

I had to look back in the file history to see what this was all about because I made these changes back in November! The problem was that in the Select Books dialog, there were a bunch of places where we were calling GetParatextScrTextWrapperIfNeeded and then blindly assuming that it would have succeeded in setting m_paratextScrTextWrapper. So then the calling code would access it and throw an exception. However, the focus of this dialog is not on getting updated content from Paratext, so we don't want to bother the user if that isn't possible. (If it is possible, then we ask them about including the updated content (when appropriate). See if the changes make sense in light of that explanation. If not, we can go through them interactively. If you think this explanation (or some modified version of it) belong as a comment in the code, I could add a summary comment on GetParatextScrTextWrapperIfNeeded.
</blockquote></details>

I added a comment.

PR #555, opened 2019-11-12
PG-1289: Prevented inconsistent state in TryLoadBlock

tombogle:
This prevents a crash if the user attempts to navigate to the next relevant block, after failing to load a block based on a reference that corresponds to chapter 0 or verse 0.
Also put in a safeguard against an invalid VerseRef (though currently this can't happen in production code).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/555)

### DIFF
--- a/Glyssen/BlockNavigator.cs
+++ b/Glyssen/BlockNavigator.cs
@@ -86,6 +86,8 @@ namespace Glyssen
 
 		public BookBlockIndices GetIndicesOfFirstBlockAtReference(VerseRef verseRef, bool allowMidQuoteBlock = false)
 		{
+			if (!verseRef.Valid)
+				return null;
 			var bookId = verseRef.Book;
 			int bookIndex = -1;
 			BookScript book = null;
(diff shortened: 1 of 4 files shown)


tombogle:
___
*[Glyssen/Dialogs/BlockNavigatorViewModel.cs, line 780 at r1](https://reviewable.io/reviews/sillsdev/glyssen/555#-LtWQoYb9J2oEFJg8Zp3:-LtZwHWbE0ECCcMtyvHm:b2ov6w4) ([raw file](https://github.com/sillsdev/glyssen/blob/3b7f017d4cdfabd79ca7419e6d0f16f838da1b9f/Glyssen/Dialogs/BlockNavigatorViewModel.cs#L780)):*

Does it matter that this message says index out of range for an argument out of range exception? 
I.e. do you want to distinguish between them for some reason?
</blockquote></details>

Probably not. Logically it sure feels like it should be an index, rather than an argument: https://stackoverflow.com/questions/22856801/index-out-of-range-exception-vs-argument-out-of-range-exception. I'm not entirely sure I agree with the decision to treat these two situations distinctly, at least in this most common case.

PR #557, opened 2019-11-13
PG-1286, PG-1291: Fixed reference text joining code to clone before joining

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/557)

### DIFF
--- a/Glyssen/BookScript.cs
+++ b/Glyssen/BookScript.cs
@@ -138,7 +138,7 @@ namespace Glyssen
 					if (!clone.CharacterIsStandard)
 						clone.CharacterIdInScript = narrator;
 				};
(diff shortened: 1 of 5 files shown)


tombogle:
The bug fix is complete. Now considering whether it is feasible and worthwhile to do a data migration to try to patch up data.

tombogle:
___
*[Glyssen/BookScript.cs, line 203 at r2](https://reviewable.io/reviews/sillsdev/glyssen/557#-LtabB6iEhhLv77AcKc3:-LtaprBh6HwhPiaV6T4i:b-j688ks) ([raw file](https://github.com/sillsdev/glyssen/blob/bfb22638462befad1e2904027c0e02a4b5c48464/Glyssen/BookScript.cs#L203)):*

To verify, you didn't change any logic in this file in the second commit, right?
</blockquote></details>

I hope not. It should have just been a straight refactoring, including changing the variable names and their order.

PR #558, opened 2019-11-15
PG-1290 Reworked the way we filter and display progress in rainbow mode

tombogle:
Now always filter and count matchups instead of blocks, regardless of filter.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/558)


tombogle:
Yeah, I hear you. This one proved to be way more than I was expecting.  Besides fixing the bug, I hope it will make the program more sensible and stable (if not in the short run, at least once the dust settles).

tombogle:
___
*[Glyssen/BookBlockIndices.cs, line 43 at r1](https://reviewable.io/reviews/sillsdev/glyssen/558#-Lu4OrxhAC8_Lp-jAckQ:-Lu5arkyCtnKP3dxRzzB:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/cc671a9e524b696e6736766c10cb4bbc3e04161f/Glyssen/BookBlockIndices.cs#L43)):*

missing closing parenthesis (if you're making changes anyway)
</blockquote></details>

Done.

tombogle:
___
*[Glyssen/Dialogs/BlockNavigatorViewModel.cs, line 181 at r1](https://reviewable.io/reviews/sillsdev/glyssen/558#-Lu4JImjAq3qFPluqzsb:-Lu5b7Kp6qazDWHxJTiA:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/cc671a9e524b696e6736766c10cb4bbc3e04161f/Glyssen/Dialogs/BlockNavigatorViewModel.cs#L181)):*

Probably worth removing or adding comments
</blockquote></details>

Done.

tombogle:
___
*[GlyssenTests/Dialogs/BlockNavigatorViewModelTests.cs, line 1486 at r1](https://reviewable.io/reviews/sillsdev/glyssen/558#-Lu43Heg7F5YV_VDZlIl:-Lu5bQUj4Jr-LuO8-chI:bfdmpqy) ([raw file](https://github.com/sillsdev/glyssen/blob/cc671a9e524b696e6736766c10cb4bbc3e04161f/GlyssenTests/Dialogs/BlockNavigatorViewModelTests.cs#L1486)):*

I'm finding it impossible to follow the logic of why certain calls to AttemptRefBlockMatchup = X became setMode(mode, !X) and some went away entirely. That's probably ok. But maybe you want to look at them again yourself to make sure they make sense to you.
</blockquote></details>

Many/most of the ones that went away were either not really needed in the first place or no longer needed because the FindRefInMark test method now works correctly with matchups. I'll have another look at the ones that got negated. I don't remember doing it, but maybe there was a logical reason.

tombogle:
___
*[GlyssenTests/Dialogs/BlockNavigatorViewModelTests.cs, line 1486 at r1](https://reviewable.io/reviews/sillsdev/glyssen/558#-Lu43Heg7F5YV_VDZlIl:-Lu5eTbh0w0NcxgsLzs5:b5tvfqp) ([raw file](https://github.com/sillsdev/glyssen/blob/cc671a9e524b696e6736766c10cb4bbc3e04161f/GlyssenTests/Dialogs/BlockNavigatorViewModelTests.cs#L1486)):*

Many/most of the ones that went away were either not really needed in the first place or no longer needed because the FindRefInMark test method now works correctly with matchups. I'll have another look at the ones that got negated. I don't remember doing it, but maybe there was a logical reason.
</blockquote></details>

You probably figured this out, but I made the decision to eliminate the direct public setters for those two properties because they both needed to cause the filter to be recomputed and that's slow enough that I didn't want to incur that cost every time both of them needed to be set. Some of the ones where we used to explicitly set it to true went away because the mode was set (in the constructor) to NotAlignedToReferenceText, which forces the property to true. Some of the ones where it was being set to false (e.g., line 987 of AssignCharacterViewModelTests.cs) were not needed because that is already the default. I probably wrote the test that way to make it more explicit (or maybe because that line was copied form another test). The places I found that kind of looked like setting it to the opposite of how it was set previously (SetBlockMatchupForCurrentVerse_VerseHasOnlyOneBlock_CurrentReferenceTextMatchupHasOneBlock, TryLoadBlock_ReferenceTextCausesSplitInVernacular_SplitBlocksAddedToNumberOfBlocksAndRelevantBlocks, etc.) were places where it used to set it to false for the setup and then true for the SUT, but I simplified/changed the setup. Some of the changes were related to making it so the navigator's (multi-block) indices and and the matchup ranges now stay in sync so that LoadNextRelevantBlock can work in a straightforward and sensible way.
Going through and reviewing them all was worthwhile. I found a few usages that were redundant and a few other places where I could delete commented out lines and make other minor refactoring improvements.

PR #573, opened 2019-12-04
GlyssenEngine Migration 3

AndrewChristensenFCBH:
Moved CharacterDetail, CharacterDetailData, and VoiceActor folder. 
Split the file CharacterGroup into 2 and called the new file CharacterGroupSupport. Moved the new file to GlyssenEngine.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/573)


AndrewChristensenFCBH:
> 
> 
> > Reviewed 47 of 47 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/573)_ status: all files reviewed, 2 unresolved discussions (waiting on @AndrewChristensenFCBH)
> 
> _[GlyssenEngine/Character/CharacterGroupSupport.cs, line 106 at r1](https://reviewable.io/reviews/sillsdev/glyssen/573#-LvI16Oe4zxoXMfPyOLN:-LvI16Oe4zxoXMfPyOLO:bckisyx) ([raw file](https://github.com/sillsdev/glyssen/blob/aaf09ec641ac7d787cd1a554723c44430c6f1835/GlyssenEngine/Character/CharacterGroupSupport.cs#L106)):_
> 
> > ```
> > CharacterVerseData.GetCharacterNameForUi
> > ```
> 
> Out of curiosity, is this what is preventing you from moving the file as is?
> 
> _[GlyssenEngine/Resources/CharacterDetail.txt, line 1 at r1](https://reviewable.io/reviews/sillsdev/glyssen/573#-LvI07hf2MtOPto4DXox:-LvI07hf2MtOPto4DXoy:b-3k1ry6) ([raw file](https://github.com/sillsdev/glyssen/blob/aaf09ec641ac7d787cd1a554723c44430c6f1835/GlyssenEngine/Resources/CharacterDetail.txt#L1)):_
> 
> > ```
> > #Character ID	Max Speakers	Gender	Age	Status	Comment	Reference	FCBH Character
> > ```
> 
> This should be a rename.
> In other words, we want to delete the file from Glyssen/Resources.

What remained in the file has a class that depends on other classes that haven't been moved yet. Block is one of them.  

Can't delete CharacterDetail (or move it) because it is still used in other locations in Glyssen.

AndrewChristensenFCBH:
> 
> 
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/573)_ status: all files reviewed, 1 unresolved discussion (waiting on @AndrewChristensenFCBH)
> 
> _[GlyssenEngine/Resources/CharacterDetail.txt, line 1 at r1](https://reviewable.io/reviews/sillsdev/glyssen/573#-LvI07hf2MtOPto4DXox:-LvIANSL-1-li6rQ0D4f:b-8e9gsk) ([raw file](https://github.com/sillsdev/glyssen/blob/aaf09ec641ac7d787cd1a554723c44430c6f1835/GlyssenEngine/Resources/CharacterDetail.txt#L1)):_
> Previously, AndrewChristensenFCBH (Andrew Christensen) wrote…
> 
> CharacterDetail.txt is still referenced in Glyssen? Why can't you point those references here?
> 
> I don't think we can have this (two copies of the file) as an in-between state.
> i.e. Which is right? Which one should you change if you need to make a change?

I had talked to Tom about this. He mentioned that we would need a master to change and then a the change would be copied over to the other. It is used 427 times in Glyssen so I don't think it will be possible to move small chunks of the code. Please talk it over with Tom and give me direction.

AndrewChristensenFCBH:
> 
> 
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/573)_ status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)
> 
> _[GlyssenEngine/Resources/CharacterDetail.txt, line 1 at r1](https://reviewable.io/reviews/sillsdev/glyssen/573#-LvI07hf2MtOPto4DXox:-LvIIn4-2iG4Jcdsc-zN:b-aark0e) ([raw file](https://github.com/sillsdev/glyssen/blob/aaf09ec641ac7d787cd1a554723c44430c6f1835/GlyssenEngine/Resources/CharacterDetail.txt#L1)):_
> Previously, AndrewChristensenFCBH (Andrew Christensen) wrote…
> 
> I'll let @tombogle weigh in, but that doesn't make sense to me. The only reference to CharacterDetail.txt (vs CharacterDetail.cs or the class) should be as "Resources.CharacterDetail". You've already moved the relevant one in CharacterDetailData. The only other two I see are in ControlDataIntegrityTests which should (I think) be able to point to the new location in GlyssenEngine.

OK, reviewing it again I see that I gave you wrong numbers, that's what I get for rushing my answer, sorry.  I made the change you requested and checked it in so it now shows it as moved instead of added.  I had deleted that file earlier and got a lot of errors (hundreds). I must have been confused.  Sometimes the compiler gives errors in waves, so maybe that was the problem.  Sorry about that.  

So besides deleting the file I also modified Resources.CharacterDetail to GlyssenEngine.Resources.CharacterDetail in 2 places in CharacterDetailDataTests.cs.
I also changed 2 access modifiers from internal to public in the Resources file.

AndrewChristensenFCBH:
> InternalsVisibleTo

That method no longer works for .NetStandard since it doesn't have an Assembly file. But I added the change to the project page and reverted all of my past changes from internal to public.

AndrewChristensenFCBH:
> 
> 
> > Reviewed 5 of 5 files at r4.
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/573)_ status: :shipit: complete! all files reviewed, all discussions resolved
> 
> _[GlyssenEngine/GlyssenEngine.csproj, line 71 at r4](https://reviewable.io/reviews/sillsdev/glyssen/573#-LvMP5WaEJPmhSf_EejC:-LvMP5WaEJPmhSf_EejD:b-g88mi2) ([raw file](https://github.com/sillsdev/glyssen/blob/10bb55eb15cd0738e0a3b3e97dedd960d844a2cc/GlyssenEngine/GlyssenEngine.csproj#L71)):_
> 
> > ```
> >     </AssemblyAttribute>
> >     <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
> >       <_Parameter1>Glyssen</_Parameter1>
> > ```
> 
> I'm thinking we don't want this (the Glyssen one) longterm. But I'm not going to hold up this change any more for that.

I don't think we will need this change long term but for now we do (unless I make something public).

PR #575, opened 2019-12-05
Moved RelatedCharactersTests under GlyssenEngineTests.

tombogle:
Note: this change required GlyssenTests to reference GlyssenEngineTests and have TestRelatedCharacters resource be public because some classes are not yet moved from GlyssenTests to GlyssenEngineTests.

Added CharacterVerseDataTests.
Renamed existing CharacterVerseDataTests to ControlCharacterVerseDataTests.
Added ControlDataIntegrityTest to ensure that extra-biblical character IDs (however we define them) are not used in CV control file.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/575)


tombogle:
___
*[GlyssenEngineTests/Properties/Resources.Designer.cs, line 25 at r1](https://reviewable.io/reviews/sillsdev/glyssen/575#-LvLpaZV8LhVpRcqbveO:-LvN3pwnD1-9sMHOsKLk:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/6444eadfb4c5b51f0cf8a8d2be4ed06c80d1a6bf/GlyssenEngineTests/Properties/Resources.Designer.cs#L25)):*

I believe Resources classes are typically internal. I'm guessing you had to manually make this public after the code was generated?
As I pointed out to Andrew in his PR, we typically do such a thing for consuming test projects by rather making internals visible to that test project.
Since this is itself a test project, I guess it *really* doesn't matter (as opposed to likely not mattering)... So up to you if it is worth the change.
</blockquote></details>

Done.

PR #578, opened 2019-12-09
Improved performance/accuracy in CharacterVerseData methods

tombogle:
Methods improved: IsCharacterOfType, IsCharacterStandard, IsCharacterExtraBiblical

Also removed a garbage comment in CharacterVerseDataTests, fixed a test name, and added a few additional test cases (that now pass because of improved implementation)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/578)

### DIFF
--- a/GlyssenEngineTests/Character/CharacterVerseDataTests.cs
+++ b/GlyssenEngineTests/Character/CharacterVerseDataTests.cs
@@ -3,23 +3,20 @@ using NUnit.Framework;
 
 namespace GlyssenEngineTests.Character
 {
-	/// <summary>
-	/// Note that these tests purport to test the GetCharacters method, but in fact that is just a simple LINQ statement;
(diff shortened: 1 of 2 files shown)


GlyssenEngineTests/Character/CharacterVerseDataTests.cs
```diff
@@ -41,9 +38,73 @@ public void IsCharacterStandard_StandardCharacter_ReturnsTrue(string characterId
 		[TestCase("narrator's friend")]
 		[TestCase("extraordinary dude in MAT")]
 		[TestCase("someone speaking in intro-REV")]
```
AndrewChristensenFCBH: These test cases were for normal characters aren't they?  They are not being used for non standard characters.  Is this correct?  This includes line 37 (Jesus).

PR #581, opened 2019-12-11
PG-1304: Bumped all .Net Framework-based DLLs to 4.7.2

tombogle:
Also added package reference to System.ValueTuple to all to avoid conflict warnings and runtime failure.
Removed System.Runtime.InteropServices.RuntimeInformation.dll form Installer (apparently a local copy is no longer needed).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/581)

### DIFF
--- a/ControlDataIntegrityTests/App.config
+++ b/ControlDataIntegrityTests/App.config
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <startup> 
-    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1" />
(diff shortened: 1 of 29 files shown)


andrew-polk:
You've evaluated the impact of leaving some versions of Windows behind and gotten buy-in from anyone necessary, I guess?

tombogle:
I checked with Bryan and he agreed that upgrading to 4.7.2 and leaving behind the archaic OS versions was okay. Yes, it's possible that adding System.ValueTuple to all the framework-based DLLs would have fixed it. But since the "nuget" branch requires us to change to 4.7.2 anyway (no, I don't remember why), it's probably not worth the effort of experimenting with that.

PR #582, opened 2019-12-16
PG-1305: Crash (no green screen): DataGridViewComboBoxCell value is not valid.

tombogle:
Fixed crash when saved mode was not a legal combination (not related to bug report, but I got this error in Sample project while investigating bug)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/582)


tombogle:
___
*[Glyssen/Dialogs/AssignCharacterDlg.cs, line 296 at r1](https://reviewable.io/reviews/sillsdev/glyssen/582#-LwEDyN57gReMKusq62r:-LwEjFBx04yXqc3k6lx6:b-6ud0dm) ([raw file](https://github.com/sillsdev/glyssen/blob/7f8d4b6f9a8b233645d7eb40e77720de298eb245/Glyssen/Dialogs/AssignCharacterDlg.cs#L296)):*

I don't understand this condition. Can the mode even equal ExcludeUserConfirmed? Is this supposed to be `&`? or a check against `> 0`?
</blockquote></details>

In the past (presumably when we had a separate thing on the toolbar for ExcludeUserConfirmed), it was possible to have both the ExcludeUserConfirmed flag and NotAlignedToReferenceText flags set, even though that now seems like a nonsensical combination. So this check just says, if your .glyssen file had that illegal combination set, assume what they really want is NotAlignedToReferenceText. But if that's not available, then fall back to the first choice.

tombogle:
___
*[Glyssen/Dialogs/AssignCharacterDlg.cs, line 296 at r1](https://reviewable.io/reviews/sillsdev/glyssen/582#-LwEDyN57gReMKusq62r:-LwEqMtI-CazZJ8qFdKL:bgpw0ra) ([raw file](https://github.com/sillsdev/glyssen/blob/7f8d4b6f9a8b233645d7eb40e77720de298eb245/Glyssen/Dialogs/AssignCharacterDlg.cs#L296)):*

So I think you want a `&` rather than a `|`
</blockquote></details>

No because a bitwise & when building a matchable flag will only turn on a bit if both things have that bit on. A bitwise | will turn on the bit if either of them have it on. And besides, it must be correct because it got past the crash!

tombogle:
___
*[Glyssen/Dialogs/AssignCharacterDlg.cs, line 296 at r1](https://reviewable.io/reviews/sillsdev/glyssen/582#-LwEDyN57gReMKusq62r:-LwEr5Rp-WdK5TfoRpv4:b-wc19ud) ([raw file](https://github.com/sillsdev/glyssen/blob/7f8d4b6f9a8b233645d7eb40e77720de298eb245/Glyssen/Dialogs/AssignCharacterDlg.cs#L296)):*

No because a bitwise & when building a matchable flag will only turn on a bit if both things have that bit on. A bitwise | will turn on the bit if either of them have it on. And besides, it must be correct because it got past the crash!
</blockquote></details>

If could have done it this way: if ((mode & BlocksToDisplay.ExcludeUserConfrimed) > 0 && (mode & BlocksToDisplay.NotAlignedToReferenceText))

tombogle:
___
*[Glyssen/Dialogs/AssignCharacterDlg.cs, line 296 at r1](https://reviewable.io/reviews/sillsdev/glyssen/582#-LwEDyN57gReMKusq62r:-LwErS8t8YmBcL_lVSBa:b-38pl29) ([raw file](https://github.com/sillsdev/glyssen/blob/7f8d4b6f9a8b233645d7eb40e77720de298eb245/Glyssen/Dialogs/AssignCharacterDlg.cs#L296)):*

You're right. Sorry.
</blockquote></details>

Though that wouldn't be exactly equivalent because it would produce a true value if any additional flags
 were also set.

tombogle:
___
*[DistFiles/reference_texts/English/NUM.xml, line 2653 at r2](https://reviewable.io/reviews/sillsdev/glyssen/582#-LwF89C11jmuWtJzLty-:-LwFL-BO5Gnc_wAkGChB:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/8e33fe762e856935396505a5e31b687ca030608e/DistFiles/reference_texts/English/NUM.xml#L2653)):*

Now there are two verse 45 markers
</blockquote></details>

Done.

tombogle:
___
*[Glyssen/Dialogs/AssignCharacterDlg.cs, line 116 at r1](https://reviewable.io/reviews/sillsdev/glyssen/582#-LwErpxXDzsDlWFy41Ji:-LwFL77G9di3FwFMrUIg:b2nhswy) ([raw file](https://github.com/sillsdev/glyssen/blob/7f8d4b6f9a8b233645d7eb40e77720de298eb245/Glyssen/Dialogs/AssignCharacterDlg.cs#L116)):*

So should this be `&`?
</blockquote></details>

Yes

PR #584, opened 2019-12-19
Removed MessageBox from all code that will be moving to GlyssenEngine

AndrewChristensenFCBH:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/584)

### DIFF
--- a/Glyssen/DataMigrator.cs
+++ b/Glyssen/DataMigrator.cs
@@ -4,11 +4,11 @@ using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
-using System.Windows.Forms;
 using DesktopAnalytics;
 using Glyssen.Bundle;
 using Glyssen.Properties;
 using Glyssen.Shared;
+using GlyssenEngine.Utilities;
 using L10NSharp;
 using SIL;
 using SIL.DblBundle;
@@ -139,7 +139,7 @@ namespace Glyssen
 													"Versification file: {2}\r\n" +
 													"Error: {3}"),
 												projectFilePath, origBundlePath, versificationPath, ex.Message);
-											MessageBox.Show(msg, GlyssenInfo.kProduct, MessageBoxButtons.OK, MessageBoxIcon.Warning);
+											MessageModal.Show(msg, GlyssenInfo.kProduct, Buttons.OK, Icon.Warning);
 											File.WriteAllText(errorlogPath, msg);
 										}
 									}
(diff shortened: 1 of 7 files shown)


AndrewChristensenFCBH:
> ```smalltalk
> WinFormsMessageModal 
> ```

All the changes you requested have been made. Some MessageBox instances occur during error or warning conditions but do not display the corresponding icons. I would suggest that these are reviewed in the future.

AndrewChristensenFCBH:
> ```smalltalk
> WinFormsMessageModal 
> ```

All the changes you requested have been made. Some MessageBox instances occur during error or warning conditions but do not display the corresponding icons. I would suggest that these are reviewed in the future.

> 
> 
> As you noted in your message to me on Slack, the original implementation you had with specific methods for warning, error, retry, etc. had a lot of advantages. I think you were right to move that direction since it will make for a cleaner implementation for non-Glyssen clients.
> However, this implementation is safer in that it is a one-to-one replacement for the Winforms MessageBox.
> My suggestion: we go ahead and merge what you have here so you aren't held up, and you can make a note for future improvement later. (Notes to your self on future improvements do not necessarily need to be made by you, but having a list will be helpful for whomever can do them!)
> 
> > Reviewed 8 of 8 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/584#-:-LwZ2nqKEM6OPSmEnMsW:b-3makmt)_ status: all files reviewed, 1 unresolved discussion (waiting on @AndrewChristensenFCBH)
> 
> _[Glyssen/Dialogs/WinFormsMessageBox.cs, line 70 at r2](https://reviewable.io/reviews/sillsdev/glyssen/584#-LwZ2UjA8qHuY1oVsGck:-LwZ2UjA8qHuY1oVsGcl:bvxa01h) ([raw file](https://github.com/sillsdev/glyssen/blob/8e0d3d7bc032cfbac398f287ac36ca169ac6590c/Glyssen/Dialogs/WinFormsMessageBox.cs#L70)):_
> 
> > ```smalltalk
> > 				return MessageResult.Retry;
> > 			// if (result == DialogResult.Yes)
> > 			return MessageResult.Yes;
> > ```
> 
> In my opinion, these would all be cleaner as switch statements.
> But maybe that's just a stylistic preference? What do you think?
> I won't hold up the PR for it if you want to just leave as is.

The problem with the original implementation is that I made a lot of assumptions. Those might not always be correct. I can make a note to go back later.

Agreed. Change made.

PR #589, opened 2020-01-08
Glyssen engine migration 8

AndrewChristensenFCBH:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/589)


AndrewChristensenFCBH:
> ```smalltalk
> fontInstallationAttempted 
> ```

All of your requests have been done. Good catch on that cleanup. I did not remove the fontInstallationAttempted parameter because it would just be pushing the test outside of the function, causing more work for development. I didn't see any benefit to doing it but I will if still think it is a better solution.

PR #596, opened 2020-01-15
Refactoring to remove Block's dependency on Glyssen.Dialogs.

tombogle:
Renamed CharacterSelect to GetHtmlSelectCodeForDropdown and broke into two separate methods with a private helper.
Removed useless code.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/596)


tombogle:
[WIP] Just tested it out and something is wrong with the drop-down display.

PR #599, opened 2020-01-16
PG-1298: Prevent orphaned trailing punctuation in USX parser

tombogle:
If USX has punctuation following a character style (\qt or \wj) that gets broken off into a separate block and that punctuation appears to belong to the preceding block, append it.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/599)


tombogle:
Yes, I think you're right. We could live without it, but if any existing project is waiting for this fix, it won't really do any good unless we "force" an update.

PR #601, opened 2020-01-22
Fixed flaw in MigrateDeprecatedCharacterIds + misc. improvements (comments, tests)

tombogle:
Moved Fonts.Default initialization into static constructor for TestPr…ject so that all unit tests that rely on that can be run independently and not depend on the order of tests.

Added early return in MigrateProjectData to avoid logging the no-op case when we're already on the correct version. Note: This means that MigrateDeprecatedCharacterIds will NOT be called if the version is already up-to-date. (I think this should be okay.)
Prevented overwriting a "Needs Review" character assignment in MigrateDeprecatedCharacterIds. (There is some special logic in the parser to assign this when blocks include verses with both implicit and non-implicit characters, and the logic in this method was clobbering those.)
Fixed typos in comments in BlockNavigatorViewModelTests.cs and Block.cs

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/601)

### DIFF
--- a/Glyssen/Block.cs
+++ b/Glyssen/Block.cs
@@ -624,7 +624,7 @@ namespace Glyssen
 		/// This handles the common case where the first Block Element is a Verse and the more
 		/// unusual case where there is a preceding Script Text consisting only of an opening square
 		/// bracket (which indicates a verse that is often omitted because of weak manuscript evidence).
-		/// Technically, any preceding SciptText element that consists entirely of punctuation will be
+		/// Technically, any preceding ScriptText element that consists entirely of punctuation will be
 		/// considered as being part of the following verse.)
 		/// </summary>
 		public bool StartsAtVerseStart
--- a/Glyssen/ProjectDataMigrator.cs
+++ b/Glyssen/ProjectDataMigrator.cs
@@ -28,6 +28,7 @@ namespace Glyssen
 			NoOp,
 			Partial,
 			Complete,
+			UpToDate,
 		}
 
 		public static MigrationResult MigrateProjectData(Project project, int fromControlFileVersion)
@@ -38,10 +39,13 @@ namespace Glyssen
 			if (!project.Books.Any())
 				return MigrationResult.NoOp;
 
-			Logger.WriteEvent($"Migrating project {project.ProjectFilePath} from {fromControlFileVersion} to {ControlCharacterVerseData.Singleton.ControlFileVersion}");
+			if (fromControlFileVersion == ControlCharacterVerseData.Singleton.ControlFileVersion)
+				return MigrationResult.UpToDate;
 
 			if (s_lastProjectMigrated != project)
 				s_migrationsRun = 0;
+
+			Logger.WriteEvent($"Migrating project {project.ProjectFilePath} from {fromControlFileVersion} to {ControlCharacterVerseData.Singleton.ControlFileVersion}");
 
 			var result = MigrationResult.Partial;
 
@@ -240,6 +244,7 @@ namespace Glyssen
 
 				foreach (Block block in book.GetScriptBlocks().Where(block => block.CharacterId != null &&
 					block.CharacterId != CharacterVerseData.kUnexpectedCharacter &&
+					block.CharacterId != CharacterVerseData.kNeedsReview &&
 					!CharacterVerseData.IsCharacterStandard(block.CharacterId)))
 				{
 					if (block.CharacterId == CharacterVerseData.kAmbiguousCharacter)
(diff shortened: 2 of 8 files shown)


tombogle:
___
*[Glyssen/ProjectDataMigrator.cs, line 43 at r1](https://reviewable.io/reviews/sillsdev/glyssen/601#-LzCxPOR0K5mwaDzHU6k:-LzDCOJJ9CjAxqnLK8Nu:b-4kdmix) ([raw file](https://github.com/sillsdev/glyssen/blob/e0d6ed136742b802be1dc47969474ad4d5d9d685/Glyssen/ProjectDataMigrator.cs#L43)):*

This seems safe, but I thought we had things set up such that we would never call MigrateProjectData if the version number was up to date.

So, it might be worth a quick check to make sure we didn't break something there.  Otherwise, :lgtm: 
</blockquote></details>

I caught it because I saw a log file where we said we were migrating from 153 to 153. I don't see anything to suggest that we ever did that check. (Blame shows it was reworked a little last March, but this aspect of the logic was fundamentally unchanged.) Originally, I think we thought this would be a quick no-op if we were up-to-date, but eventually we decided we always needed to run MigrateDeprecatedCharacterIds. I can only think/hope that we did that because we believed (at the time) that this would only get run if the version numbers were not equal. The only other reason would be to perform constant "sanity" cleanup to deal with program bugs. (There is, for example, a bit of code at the start of the main loop to set UserConfirmed to false if the character ID is Ambiguous. That code seemingly has nothing to do with the CV control file, so that might be needed no matter what. I hope not though.)

tombogle:
___
*[Glyssen/ProjectDataMigrator.cs, line 43 at r1](https://reviewable.io/reviews/sillsdev/glyssen/601#-LzCxPOR0K5mwaDzHU6k:-LzDFhfS-snkeOwZr8Ni:bn9zk09) ([raw file](https://github.com/sillsdev/glyssen/blob/e0d6ed136742b802be1dc47969474ad4d5d9d685/Glyssen/ProjectDataMigrator.cs#L43)):*

I caught it because I saw a log file where we said we were migrating from 153 to 153. I don't see anything to suggest that we ever did that check. (Blame shows it was reworked a little last March, but this aspect of the logic was fundamentally unchanged.) Originally, I think we thought this would be a quick no-op if we were up-to-date, but eventually we decided we always needed to run MigrateDeprecatedCharacterIds. I can only think/hope that we did that because we believed (at the time) that this would only get run if the version numbers were not equal. The only other reason would be to perform constant "sanity" cleanup to deal with program bugs. (There is, for example, a bit of code at the start of the main loop to set UserConfirmed to false if the character ID is Ambiguous. That code seemingly has nothing to do with the CV control file, so that might be needed no matter what. I hope not though.)
</blockquote></details>

That bit of code (referred to in parentheses at the end of my long comment above) was added 9/28/2016 with no reference to a Jira #. Just says: "Fixed data migration problem that could cause a block to be marked as Ambiguous and user-confirmed."

tombogle:
___
*[Glyssen/ProjectDataMigrator.cs, line 43 at r1](https://reviewable.io/reviews/sillsdev/glyssen/601#-LzCxPOR0K5mwaDzHU6k:-LzDGj685xPwrx1eOiSd:b-hi852t) ([raw file](https://github.com/sillsdev/glyssen/blob/e0d6ed136742b802be1dc47969474ad4d5d9d685/Glyssen/ProjectDataMigrator.cs#L43)):*

That bit of code (referred to in parentheses at the end of my long comment above) was added 9/28/2016 with no reference to a Jira #. Just says: "Fixed data migration problem that could cause a block to be marked as Ambiguous and user-confirmed."
</blockquote></details>

In the accompanying test, it says the scenario was caused by ba bug in a previous version of the method. So really the fix should have been tied to a version number, rather than being done every time. Maybe we did that to save an additional step requiring us to run through the data set, but in the long run, it means doing it every time for years after that bug is gone.
So then the only remaining reason I could see for doing this every time would be to catch cases where we make changes but fail to update the version number. (That could theoretically still happen when developing, but we seem to have gotten pretty disciplined about that for releases.)

tombogle:
___
*[Glyssen/ProjectDataMigrator.cs, line 43 at r1](https://reviewable.io/reviews/sillsdev/glyssen/601#-LzCxPOR0K5mwaDzHU6k:-LzE4QTCC8FxDqWI2FAI:b3mfpat) ([raw file](https://github.com/sillsdev/glyssen/blob/e0d6ed136742b802be1dc47969474ad4d5d9d685/Glyssen/ProjectDataMigrator.cs#L43)):*

Ok, don't hold it up for me. If you're comfortable, I'm comfortable.
</blockquote></details>

I'm never comfortable. I have you review stuff so I can blame you later when it's wrong.

PR #603, opened 2020-01-22
PG-1315: Fix crash when vernacular verse maps into multi verse ref block in prev chapter

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/603)


tombogle:
WIP: Also need to figure out how to trigger a re-parse when the versification changes (in Paratext).

PR #605, opened 2020-01-23
PG-1275: Include single-block matchups in the Needs Alignment filter 

tombogle:
when the block has a different character from the corresponding reference text block

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/605)


tombogle:
WIP: Waiting for Ken Steffen to weigh in on the question of how best to address this.

PR #608, opened 2020-01-28
Glyssen engine migration - big monster

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/608)


tombogle:
Bug fixes included in big monster:
* corrected parser version number constant to match the previous value stored in the Glyssen setting
* Refactoring to fix BlockTests dependency on the order that tests are run (this was a bug, but we were getting lucky that it wasn't failing)

Refactoring/renaming that could (probably) be done as separate PRs:
* ScriptBlocksViewType into separate file
* Localizer -> LocalizationManager (inside Glyssen)
* Added folders/namespaces for improved organization: Script, Paratext, UndoActions
* Move of GetAllBlocksWhichContinueTheQuoteStartedByBlock and GetIndicesOfQuoteContinuationBlocks to BlockNavigator (to simplify BookScript).
* Reducing the options in MessageModal to the minimum set required.
* In theory, moving all the unit tests could probably be done separately, but I'm not sure this is logical or ideal. (In any case, almost all of the changes will be trivial to review.)
Renaming and minimizing IAnalytics to IErrorHandler and coalescing GlyssenEngine.Utilities.Analytics and NullAnalytics into static NonFatalErrorHandler. (Not sure how easily this could be pulled into separate PR because of how I used this to do the needed refactoring in Project.)

Other improvements (not related to GE refactoring):
* Capitalized Text Release Bundle consistently (treating it as a proper noun)

tombogle:
___
*[GlyssenEngine/GlyssenEngine.csproj, line 71 at r5](https://reviewable.io/reviews/sillsdev/glyssen/608#-M-C9H0cCRBA5C3Yg5kU:-M-CGFM3AbYjtpSYhmHA:b-7w2die) ([raw file](https://github.com/sillsdev/glyssen/blob/821dfafaeb55f9a21db120b860a78e84afc37912/GlyssenEngine/GlyssenEngine.csproj#L71)):*

Doesn't seem like a file which should be in GlyssenEngine
</blockquote></details>

And the file wasn't. Just this bizarre reference to it. Not sure why that didn't trigger some kind of build error or warning.

tombogle:
___
*[GlyssenEngine/GlyssenEngine.csproj, line 98 at r5](https://reviewable.io/reviews/sillsdev/glyssen/608#-M-C9GFr20xL5iWz1md6:-M-CGir3AB2JSgVZbs6n:bj5ssat) ([raw file](https://github.com/sillsdev/glyssen/blob/821dfafaeb55f9a21db120b860a78e84afc37912/GlyssenEngine/GlyssenEngine.csproj#L98)):*

Doesn't seem like a file which should be in GlyssenEngine
</blockquote></details>

It's probably not ever really going to be needed in practice. But there is currently an export option  that makes copies of this tiny WAV file to create "blank" clips that the user can record over.

tombogle:
___
*[GlyssenEngine/GlyssenEngine.csproj, line 90 at r5](https://reviewable.io/reviews/sillsdev/glyssen/608#-M-C966S4fGIpJKvH-TI:-M-CH7op-DXO152YbrD9:bl4l583) ([raw file](https://github.com/sillsdev/glyssen/blob/821dfafaeb55f9a21db120b860a78e84afc37912/GlyssenEngine/GlyssenEngine.csproj#L90)):*

Doesn't seem like a file which should be in GlyssenEngine
</blockquote></details>

On the one hand, I would argue that everything that has to do with producing HTML belongs in the UI layer, but on the other hand, there is a fair bit of logic surrounding it that seems to fit in the Engine, and *if* they need/want to do something similar in Vessel, it could well prove useful, so at least for now, I opted to leave that in the engine. That said, the CSS really has to do with presentation, so I think you're right that really doesn't belong in the Engine. Since the element names are hard-coded in the HTML (in the Engine code) and also in the CSS, I'm trying to think what would be the best way to handle it so that there is no implicit link between the two. My current thought is to maybe have the model make the element names available as public constants so that the client can format up the CSS using them.

tombogle:
___
*[GlyssenEngine/Script/Block.cs, line 1 at r5](https://reviewable.io/reviews/sillsdev/glyssen/608#-M-CAUtVCku0cBlTnr1U:-M-FWD536R-r2IxAiovr:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/821dfafaeb55f9a21db120b860a78e84afc37912/GlyssenEngine/Script/Block.cs#L1)):*

leftover
</blockquote></details>

Done.

tombogle:
___
*[GlyssenTests/Properties/Resources.resx, line 121 at r5](https://reviewable.io/reviews/sillsdev/glyssen/608#-M-CCb1_5bh5qJMABVTo:-M-FXIFGBUY0-uqsLZBg:b-izq7vy) ([raw file](https://github.com/sillsdev/glyssen/blob/821dfafaeb55f9a21db120b860a78e84afc37912/GlyssenTests/Properties/Resources.resx#L121)):*

Does this belong in GlyssenTests?
</blockquote></details>

Not sure if it belongs anywhere. The CharacterGroupTemplate stuff is not currently used in production. It certainly could be moved to the Engine and probably should be if it is to be used at all. Alternatively, we could junk it, or move it off into a separate DLL that does not get installed. Anyway, didn't seem to be important to move it as part of this PR. (I didn't want to needlessly bloat it.)

tombogle:
___
*[GlyssenEngine/Resources.resx, line 120 at r5](https://reviewable.io/reviews/sillsdev/glyssen/608#-M-C9kcW8c7zPjUFsXsr:-M-FZERa8VvAAQwmrHNM:b-3bvhq5) ([raw file](https://github.com/sillsdev/glyssen/blob/821dfafaeb55f9a21db120b860a78e84afc37912/GlyssenEngine/Resources.resx#L120)):*

This can't be right, can it?
</blockquote></details>

I assume you mean the apparent references to System.Windows.Forms? I don't know. It's generated, and it doesn't seem to cause any problems or require any bad reference to System.Windows.Forms in the dependencies. However, according to this post, maybe this isn't really what we want: https://stackoverflow.com/questions/32249988/resource-and-references-to-system-windows-forms-in-web-applications. As long as we're running in a process that does have the ability to load System.Windows.Forms, we'll be fine, but that might not be the case for all future users of GlyssenEngine (e.g., Vessel running on Linux, Virtual Recording). So maybe we need to make these embedded resources? Or maybe we just want to install them as independent files and deal with the extra fail-over cases when a file is missing, corrupt, etc.

tombogle:
___
*[GlyssenEngine/Resources.resx, line 127 at r5](https://reviewable.io/reviews/sillsdev/glyssen/608#-M-C9dd7-dwjoN19aPeo:-M-FbjlCBLqg4pAmw_un:b-3osq0x) ([raw file](https://github.com/sillsdev/glyssen/blob/821dfafaeb55f9a21db120b860a78e84afc37912/GlyssenEngine/Resources.resx#L127)):*

Doesn't seem like a file which should be in GlyssenEngine
</blockquote></details>

See previous comment re: same file (in GlyssenEngine.csproj)

PR #611, opened 2020-01-30
Removed BookScript's dependency on BlockNavigatorViewModel…

tombogle:
… by moving two methods to BlockNavigator.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/611)

### DIFF
--- a/Glyssen/BlockNavigator.cs
+++ b/Glyssen/BlockNavigator.cs
@@ -384,11 +384,11 @@ namespace Glyssen
 			{
 				case MultiBlockQuote.Start:
 					yield return firstBlock;
-					foreach (var i in GetIndicesOfQuoteContinuationBlocks(firstBlock))
+					foreach (var i in GetIndicesOfQuoteContinuationBlocks(firstBlock, adjustmentToBlockCount))
 						yield return CurrentBook[i];
 					break;
 				case MultiBlockQuote.Continuation:
-					// These should all be brought in through a Start block, so don't do anything with them here - Should this throw an exception?
+					// These should all be brought in through a Start block, so don't do anything with them here.
 					break;
 				default:
 					// Not part of a multi-block quote. Just return the base-line block
--- a/Glyssen/BlockNavigator.cs
+++ b/Glyssen/BlockNavigator.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using GlyssenEngine;
 using SIL.Extensions;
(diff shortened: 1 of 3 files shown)


tombogle:
___
*[Glyssen/BlockNavigator.cs, line 391 at r1](https://reviewable.io/reviews/sillsdev/glyssen/611#-Lzqyjic8vppW0AIT0ed:-Lzr5PZU6jgPhtEsxXc7:b-x2lhsi) ([raw file](https://github.com/sillsdev/glyssen/blob/c75015a17759604ddc2ee40c271fa407a0422907/Glyssen/BlockNavigator.cs#L391)):*
> Should this throw an exception?
I looked at the code, and you appear to have a good memory.

tombogle:
___
*[Glyssen/BlockNavigator.cs, line 400 at r1](https://reviewable.io/reviews/sillsdev/glyssen/611#-LzqzqxSBZko8vZdOtVz:-Lzr5jpw8a5TAX10gtQ-:b-n2shgb) ([raw file](https://github.com/sillsdev/glyssen/blob/c75015a17759604ddc2ee40c271fa407a0422907/Glyssen/BlockNavigator.cs#L400)):*

> ```
> int adjustmentToBlockCount = 0
> ```
I don't understand why this is a optional parameter.
</blockquote></details>

Probably because it was optional in the above method and it felt "logically" optional, even though no existing callers omit the parameter. Everything builds with it non-optional, but I'm not sure that's really the best thing to do. I guess if we ever need to call it in a case where there is no adjustment, it would be easy enough to pass a 0 or make it optional again, but by making it optional, it sort of feels like documentation to tell our future selves that no adjustment is a pretty "normal" situation.

tombogle:
___
*[Glyssen/BlockNavigator.cs, line 400 at r1](https://reviewable.io/reviews/sillsdev/glyssen/611#-LzqzqxSBZko8vZdOtVz:-Lzr7pz879Xgr6KP2mgG:b-w6m5mg) ([raw file](https://github.com/sillsdev/glyssen/blob/c75015a17759604ddc2ee40c271fa407a0422907/Glyssen/BlockNavigator.cs#L400)):*

Probably because it was optional in the above method and it felt "logically" optional, even though no existing callers omit the parameter. Everything builds with it non-optional, but I'm not sure that's really the best thing to do. I guess if we ever need to call it in a case where there is no adjustment, it would be easy enough to pass a 0 or make it optional again, but by making it optional, it sort of feels like documentation to tell our future selves that no adjustment is a pretty "normal" situation.
</blockquote></details>

That said, obviously making it optional burned me in this case, so I could see it either way.

PR #616, opened 2020-01-30
Moved several Analytics.Track calls into the UI layer:

tombogle:
"AddCharacter", "NarratorToQuote", "SetSingleVoice", "ProjectAnalysis"
Small amount of misc code cleanup and some changes to the scope of certain class members to prepare for Glyssen Engine moves.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/616)

### DIFF
--- a/Glyssen/Character/ProjectCharacterVerseData.cs
+++ b/Glyssen/Character/ProjectCharacterVerseData.cs
@@ -27,7 +27,7 @@ namespace Glyssen.Character
 				LoadData("");
 		}
 
-		public virtual void AddEntriesFor(int bookNumber, Block block)
+		public virtual bool AddEntriesFor(int bookNumber, Block block)
 		{
 			bool added = false;
 			foreach (var verse in block.AllVerses)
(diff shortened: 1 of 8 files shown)


tombogle:
___
*[Glyssen/Dialogs/AssignCharacterDlg.cs, line 1260 at r1](https://reviewable.io/reviews/sillsdev/glyssen/616#-LzsDL2-6U590VzN2hX-:-LzsH9IX7iRQyg6HksYg:b-osd082) ([raw file](https://github.com/sillsdev/glyssen/blob/701d75c7dba5c69d5d3adf5e3cda558580ad6e3f/Glyssen/Dialogs/AssignCharacterDlg.cs#L1260)):*

SetCurrentBookSingleVoice returns early if we are not actually changing the value. Can this really happen? Or was that a just-in-case? If it can really happen, we should not send the analytics if we are not changing the value.
</blockquote></details>

Good catch. yes, it can happen. We reset the check box every time we change books.

tombogle:
___
*[Glyssen/Dialogs/ExportDlg.cs, line 251 at r1](https://reviewable.io/reviews/sillsdev/glyssen/616#-LzsAbT-7J6hDGRsYL-z:-LzsJLDz8qMrBGM5YLuL:b-1ysz2f) ([raw file](https://github.com/sillsdev/glyssen/blob/701d75c7dba5c69d5d3adf5e3cda558580ad6e3f/Glyssen/Dialogs/ExportDlg.cs#L251)):*

Seems like this should probably go inside the if block just below.
</blockquote></details>

I think we're just wanting to know that the user tried to export, even if there was ultimately some problem that thwarted them.

PR #617, opened 2020-01-30
Redid the approach to ensuring that the project font files get installed on the fly

tombogle:
The main purpose of this was to get as much of the logic into the non-UI layer and thereby support alternative approaches to the UI portion of the task for a non-Winforms environment.

Also replaced deprecated DirectoryUtilities.DeleteDirectoryRobust with RobustIO.DeleteDirectoryAndContents

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/617)


tombogle:
___
*[Glyssen/Project.cs, line 90 at r1](https://reviewable.io/reviews/sillsdev/glyssen/617#-LzsOUGI9oR7htxsmP8V:-LzsXghZ6c3SckP7-10X:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/a7086a04478540fae1891f8d5608d8e50b86be10/Glyssen/Project.cs#L90)):*

> ```
> s_fontRepository
> ```
Since this is public, I think I would call it FontRepository.
Otherwise, :lgtm:
</blockquote></details>

Done.

PR #619, opened 2020-01-31
Moved some UI-oriented methods from VoiceActorAssignmentViewModel to dialog

tombogle:
Broke up logic to build data table to keep business logic in model but put data table stuff (which is UI-specific into the Dlg). Changed tests accordingly.
Fixed bug where a localizer ID was used for two different strings.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/619)


tombogle:
Yeah, it wasn't in any of the other tmx files.

PR #626, opened 2020-02-02
Moved non-UI part of exporting a glyssenshare file into project

tombogle:
Replaced UpdatePercentInitialized() with simple property.
Made some internal classes and methods public
Misc whitespace cleanup in Project.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/626)

### DIFF
--- a/Glyssen/MainForm.cs
+++ b/Glyssen/MainForm.cs
@@ -1240,45 +1240,25 @@ namespace Glyssen
 		/// <param name="e"></param>
 		private void Export_Click(object sender, EventArgs e)
 		{
+			void HandleCustomReferenceText(string customIdentifier)
+			{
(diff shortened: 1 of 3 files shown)


tombogle:
___
*[Glyssen/MainForm.cs, line 1255 at r1](https://reviewable.io/reviews/sillsdev/glyssen/626#-M-B5GqV-rWnjT0Rko-o:-M-BRjsj5NiWag2tZlJZ:b-dft5oh) ([raw file](https://github.com/sillsdev/glyssen/blob/7bfbdc8bca649c16696cbbec17edc89c80dec8ac/Glyssen/MainForm.cs#L1255)):*

Did you intend to call ExportShare twice?
If this is necessary for some reason, we definitely need to document it.
But I'm guessing that was a mistake.
</blockquote></details>

That's the kind of mistakes that is possible after midnight.

PR #637, opened 2020-02-07
Move final Character files to the engine. Refactoring is complete.

AndrewChristensenFCBH:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/637)


GlyssenEngineTests/Character/Template/CharacterAssignerTests.cs
```diff
@@ -1,8 +1,12 @@
-﻿using System;
+﻿/// <summary>
```
tombogle: CharacterAssignerTests should not have been moved into Template. They should still be set to compile

GlyssenEngineTests/Properties/Resources.Designer.cs
```diff
@@ -378,6 +378,16 @@ internal class Resources {
             }
         }
         
+        /// <summary>
```
tombogle: Technically, there is no reason to add this as a resource, since we are not compiling the tests that would need it. It probably doesn't hurt anything, though.

GlyssenEngineTests/Character/Template/CharacterAssignerTests.cs
AndrewChristensenFCBH: Moved. Sorry, I got sloppy.

GlyssenEngineTests/Properties/Resources.Designer.cs
AndrewChristensenFCBH: I left it. It will be available if and when needed.

GlyssenEngineTests/Character/CharacterAssignerTests.cs
```diff
@@ -1,8 +1,12 @@
-﻿using System;
+﻿/// <summary>
```
tombogle: Remove this comment

AndrewChristensenFCBH: My last change. I can't make it easy for you can I?  Sorry.  Done.

PR #638, opened 2020-02-07
PG-1327: Closed up a loophole in MatchVernBlocksToReferenceTextBlocks…

tombogle:
… to prevent calling SetUnmatchedReferenceBlocks when forceMatch is passed in as true.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/638)


tombogle:
___
*[GlyssenEngine/ReferenceText.cs, line 480 at r1](https://reviewable.io/reviews/sillsdev/glyssen/638#-M-WrMNfE30iV6GPlJ5A:-M-jVLKD6p3H7E1-LKsC:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/e64a09a06521d472018ba5e452bf7f9e74f327dc/GlyssenEngine/ReferenceText.cs#L480)):*

Does this comment belong on the method? Also, how does the caller do so?
Same below.
</blockquote></details>

Done.

PR #645, opened 2020-02-14
PG-1333 [WIP] Initial idea: hack out use of System.Drawing from export

tombogle:
This solves the bug, but it is a loss of functionality. That might be okay since it's possible FCBH doesn't even care about the coloring. Other ideas:

- Include system.drawing.common.dll in Installer. (This also fixes the problem, but seems like we shouldn't need/want to do that. See: https://github.com/sillsdev/Glyssen/pull/647)

- Have callbacks to Glyssen to set the colors. (Seems like a weird thing to require the "UI" layer to handle. Haven't tested this approach. See: https://github.com/sillsdev/Glyssen/pull/648)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/645)

### DIFF
--- a/GlyssenEngine/Export/ProjectExporter.cs
+++ b/GlyssenEngine/Export/ProjectExporter.cs
@@ -2,7 +2,7 @@
 using System.Collections.Generic;
 using System.Data;
 using System.Diagnostics;
-using System.Drawing;
+//using System.Drawing;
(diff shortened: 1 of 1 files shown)


tombogle:
https://github.com/sillsdev/Glyssen/pull/648

PR #647, opened 2020-02-14
PG-1333: Added System.drawing.common.dll to Installer

tombogle:
This is alternative approach to solve PG-1333. (See also https://github.com/sillsdev/Glyssen/pull/645)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/647)


tombogle:
Opted to go with the approach in https://github.com/sillsdev/Glyssen/pull/648

PR #650, opened 2020-02-19
PG-1336: Prevent creation of matchup that splits between two blocks that end/start with same verse

tombogle:
This deals with the In the unusual case where a verse number is repeated in two adjacent blocks.

Simplified a couple BlockMatchup methods that were taking Versification. (We can get that from the book now.)
Fixed problem in unit tests where we were getting a data migration warning because we were accessing Project class before setting GlyssenInfo.Product to the test folder.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/650)

### DIFF
--- a/GlyssenEngine/Script/BlockMatchup.cs
+++ b/GlyssenEngine/Script/BlockMatchup.cs
@@ -13,17 +13,13 @@ namespace GlyssenEngine.Script
 	public class BlockMatchup
 	{
 		private readonly BookScript m_vernacularBook;
-		private readonly int m_iStartBlock;
 		private readonly PortionScript m_portion;
-		private int m_numberOfBlocksAddedBySplitting = 0;
 		private readonly IReferenceLanguageInfo m_referenceLanguageInfo;
-		private ScrVers m_versification;
 
 		public BlockMatchup(BookScript vernacularBook, int iBlock, Action<PortionScript> splitBlocks,
 			Func<Block, bool> isOkayToBreakBeforeBlock, IReferenceLanguageInfo heSaidProvider, uint predeterminedBlockCount = 0)
 		{
 			m_vernacularBook = vernacularBook;
-			int bookNum = m_vernacularBook.BookNumber;
 			m_referenceLanguageInfo = heSaidProvider;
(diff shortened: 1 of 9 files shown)


tombogle:
___
*[GlyssenEngineTests/Script/BlockMatchupTests.cs, line 133 at r1](https://reviewable.io/reviews/sillsdev/glyssen/650#-M0TUWl37yQRxgBKINS2:-M0TYnje7RmTF7_8pEnv:bsfkmst) ([raw file](https://github.com/sillsdev/glyssen/blob/7d3a01c6ebf95e9b709cb0cebc2a0047f2f0a8cf/GlyssenEngineTests/Script/BlockMatchupTests.cs#L133)):*

> ```
> trailingVerseBridgeStart
> ```
It isn't really a bridge start, right?
But then, where is the vs 12 marker?

I'm really confused on the setup. The third block claims to go from 11 to 12. It has an 11 or 11b marker and a 13 marker.
</blockquote></details>

Fixed. (The 12 does not indicate the last verse in the block, just the last verse in the initial bridge, which I failed to create).

tombogle:
___
*[GlyssenEngineTests/Script/BlockMatchupTests.cs, line 133 at r1](https://reviewable.io/reviews/sillsdev/glyssen/650#-M0TUWl37yQRxgBKINS2:-M0TaRRBA9tXG9vg6uSm:bzemgyr) ([raw file](https://github.com/sillsdev/glyssen/blob/7d3a01c6ebf95e9b709cb0cebc2a0047f2f0a8cf/GlyssenEngineTests/Script/BlockMatchupTests.cs#L133)):*

> vernacularBlocks.Add(new Block("p", 1, 11, 12)

Shouldn't the last block end in vs 13 rather than 12?
</blockquote></details>

Re-read my comment. The 12 is the InitiaiEndVerseNum. For performance reasons, we store both the initial starting number and the initial ending number because blocks don't have access to their owning list, so they would otherwise have no way to know what verse(s) they start with in the case where the first BlockElement is not a Verse. We don't need to store the last verse number because a block can easily figure that out by examining the Verse elements it holds.

PR #652, opened 2020-02-24
Reworked plumbing in GenerateCharacterGroups_IsCancelable

tombogle:
This should ensure that worker thread finishes before checking results and ending the test, preventing this test from affecting another case in the same fixture.
Also Upgraded GlyssenSharedTests to latest version of NUnit 3.

(Note: I had previously added this attribute to the one test that had begun failing on TeamCity, but I couldn't find any reason why any of these tests could safely be run in parallel, so I wanted to apply that decision to all of them and document why.)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/652)

### DIFF
--- a/GlyssenEngineTests/Rules/CharacterGroupGeneratorTests.cs
+++ b/GlyssenEngineTests/Rules/CharacterGroupGeneratorTests.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
 using GlyssenEngine;
 using GlyssenEngine.Bundle;
 using GlyssenEngine.Casting;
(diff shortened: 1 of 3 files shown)


tombogle:
___
*[GlyssenEngineTests/Rules/CharacterGroupGeneratorTests.cs, line 49 at r2](https://reviewable.io/reviews/sillsdev/glyssen/652#-M0yh5fACCaxynbH907o:-M11CvUC7jQd600xYQTB:b-oc7879) ([raw file](https://github.com/sillsdev/glyssen/blob/cab0222c078228bf1eb89aec96fea4a7ddf37dbe/GlyssenEngineTests/Rules/CharacterGroupGeneratorTests.cs#L49)):*

Having this line twice might be what is causing it to show up in the log twice. : )
Same just below.
</blockquote></details>

Yeah, I eventually spotted that and thought the same thing.

PR #653, opened 2020-03-02
Bumped up SIL.Windows.Forms to get registered trademark version of SIL logo

tombogle:
(This is the latest beta.)
Bumped up other packages to consolidate on single version
Bumped up copyright year to 2020
Bumped up to latest version of L10nSharp.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/653)


tombogle:
___
*[Glyssen/packages.config, line 44 at r1](https://reviewable.io/reviews/sillsdev/glyssen/653#-M1QyJ5q0ZvemiF0GQHr:-M1ROC9-CN7TDEeIuBnv:b-8ki0yp) ([raw file](https://github.com/sillsdev/glyssen/blob/3a9ac383e6a13a5b6feaa896f0f73fb0fa9c53b5/Glyssen/packages.config#L44)):*

There are several SIL dlls in here which aren't at the 0077. Does that just mean all the dlls on nuget don't necessarily get updated at the same time?
</blockquote></details>

No, it just means I didn't have to upgrade those to get the changes I needed. I guess there's probably no good reason not to keep them all on the same version for simplicity though.

PR #659, opened 2020-03-13
Isolate project persistence layer

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/659)


tombogle:
___
*[DevTools/AnnotationExtractor.cs, line 1 at r4](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2Kgv2b-3Fl6smRZYQK:-M2evfXk4CIisFm0z-rt:bb74x83) ([raw file](https://github.com/sillsdev/glyssen/blob/b1e695fc0d86f1a470c01b7fa74dd16d5af56682/DevTools/AnnotationExtractor.cs#L1)):*

I guess this is just cleaning up a file we already weren't using?
</blockquote></details>

True

tombogle:
___
*[Glyssen/MainForm.cs, line 49 at r2](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2JdgdJ5ImX4KewrXWH:-M2ewF1HEstnsXINqHYo:bv6nfxo) ([raw file](https://github.com/sillsdev/glyssen/blob/8f72f618f11bfe96daceafa464badb796b4f48bd/Glyssen/MainForm.cs#L49)):*

Consider putting in GlyssenShared instead? Or at least GlyssenEngine?n
</blockquote></details>

Thought about it, but it isn't used anywhere else.

tombogle:
___
*[GlyssenEngine/Project.cs, line 1664 at r2](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2Ji8dHE_vQyfnBMpdE:-M2jRwgREba49Mtoi_wT:b6gpbuk) ([raw file](https://github.com/sillsdev/glyssen/blob/8f72f618f11bfe96daceafa464badb796b4f48bd/GlyssenEngine/Project.cs#L1664)):*

> ```
> requests
> ```
requested
</blockquote></details>

See https://github.com/sillsdev/libpalaso/pull/917

tombogle:
___
*[GlyssenEngine/Project.cs, line 1666 at r2](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2JiU4e-Uhqa7LJ5WHf:-M2jS-s10XlJ1tiq7_lL:b3sk72h) ([raw file](https://github.com/sillsdev/glyssen/blob/8f72f618f11bfe96daceafa464badb796b4f48bd/GlyssenEngine/Project.cs#L1666)):*

Either way, I guess. It does make sense for this to be the default...
</blockquote></details>

We'll see what Eberhard thinks

tombogle:
___
*[GlyssenEngine/Project.cs, line 1635 at r5](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2ektnQCkrWjBEHSLPc:-M2jS3WsDJ1khKADIjyM:b6gpbuk) ([raw file](https://github.com/sillsdev/glyssen/blob/a167cce397904121a02d7c070d968f696e0f33a1/GlyssenEngine/Project.cs#L1635)):*

TODO
</blockquote></details>

See https://github.com/sillsdev/libpalaso/pull/917

tombogle:
___
*[GlyssenEngine/Project.cs, line 1664 at r2](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2Ji8dHE_vQyfnBMpdE:-M2n2ZrR8yyI7fdV74Lo:bkxwe55) ([raw file](https://github.com/sillsdev/glyssen/blob/8f72f618f11bfe96daceafa464badb796b4f48bd/GlyssenEngine/Project.cs#L1664)):*

I'm impressed if you fixed a typo in a GlyssenEngine comment with a PR in libpalaso!
</blockquote></details>

Prepare to be impressed!

tombogle:
___
*[GlyssenEngine/Project.cs, line 1679 at r2](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2JiED_E0f-7mezitBa:-M2n2hUp61w_e9ZTnS6z:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/8f72f618f11bfe96daceafa464badb796b4f48bd/GlyssenEngine/Project.cs#L1679)):*

> ```
> reader?.Dispose();
> ```
Elvis can leave the building.
(You've already checked it isn't null.)
</blockquote></details>

Done.

tombogle:
___
*[GlyssenEngine/Project.cs, line 1824 at r5](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2itdA26bB7JgLuOF6D:-M2n3U3843_mEq6JHp1s:b-hctoym) ([raw file](https://github.com/sillsdev/glyssen/blob/a167cce397904121a02d7c070d968f696e0f33a1/GlyssenEngine/Project.cs#L1824)):*
> RestoreResourceFromBackup
I wasn't entirely comfortable with saying replace because I thought it maybe suggested an implementation detail that might not be accurrate. But I like RestoreResourceFromBackup.

tombogle:
___
*[GlyssenEngine/Character/CharacterGroupList.cs, line 91 at r2](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2JilfO988Nw6jMmVcF:-M2n4bc4BwUHfbgh4D4V:b-z08mrq) ([raw file](https://github.com/sillsdev/glyssen/blob/8f72f618f11bfe96daceafa464badb796b4f48bd/GlyssenEngine/Character/CharacterGroupList.cs#L91)):*

This follows a different pattern than the VoiceActorList. There, you start by deserializing, and if the result is null, return a new list. Here, you check data is null first then assume the result is not null. Should we make them consistent?
</blockquote></details>

It's because if it's a new list, I don't need to bother with the foreach to initialize the groups. It's just a very minor optimization.

tombogle:
___
*[GlyssenEngine/Script/BookScript.cs, line 48 at r5](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2iwXD_0xY_uk7dekiL:-M2n5-Ak7oXY2IaHgVj9:b-xqh3q6) ([raw file](https://github.com/sillsdev/glyssen/blob/a167cce397904121a02d7c070d968f696e0f33a1/GlyssenEngine/Script/BookScript.cs#L48)):*

Maybe this is paranoia, but is it worth a Debug.Assert that the reader is disposed before leaving the method (since you are relegating dispose to method which doesn't obviously dispose)?
</blockquote></details>

Would be nice. Unfortunately, IDisposable does not have an IsDisposed member (as I found out when I wrote the unit test for my PR in libpalaso). But at least the unit test there shows that Dispose gets called when it's supposed to.

tombogle:
___
*[GlyssenEngineTests/ReferenceTextIdentifierTests.cs, line 93 at r3](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2KFqMHEBSFGFNNwzy-:-M2nFExw-YoYepe9BDiA:bnqnphk) ([raw file](https://github.com/sillsdev/glyssen/blob/55c97d1a9f75dbc5b7584ba5c62abedf58f8e749/GlyssenEngineTests/ReferenceTextIdentifierTests.cs#L93)):*
> ```
> #region REVIEW: Should we delete these tests? They test a method only used here.
> ```
Yes, but IsCustomReferenceAvailable is not used outside of the unit tests. So if we stopped testing that method (which is now a one-line pass-through for Reader.ResourceExists), we could get rid of it. I think the tests for AllAvailable essentially   test the internal functionality (of both the test setup and the actual production code) that we care about.

tombogle:
___
*[GlyssenEngineTests/ReferenceTextTests.cs, line 4207 at r5](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2iyElR64qTCd0Tpt_f:-M2nI8p55k3KXKwwy_rG:bjjb8kj) ([raw file](https://github.com/sillsdev/glyssen/blob/a167cce397904121a02d7c070d968f696e0f33a1/GlyssenEngineTests/ReferenceTextTests.cs#L4207)):*
> ```
> DeleteTempCustomReferenceProjectFolder()
> ```
How about CleanupUpTempImplementationAndRestorePreviousImplementation? It's some unusual plumbing, to be sure. As an alternative, I could add a property that returns the current implementation instead of having this method return it. Of the three places where this method is called, only one actually uses the return value.

tombogle:
___
*[GlyssenFileBasedPersistence/ProjectRepository.cs, line 74 at r5](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2dv9JNCTH58SmgeyZD:-M2nKYJ76v00Fif2419e:b-gh7v22) ([raw file](https://github.com/sillsdev/glyssen/blob/a167cce397904121a02d7c070d968f696e0f33a1/GlyssenFileBasedPersistence/ProjectRepository.cs#L74)):*

> bool requireWritable = false

Another case where we should force the dev to choose?
(evidence to support a yes answer is that you had to add 'true' in a subsequent commit in a test setup)
</blockquote></details>

The only instance where we pass true is in test code, so I got rid of the parameter and just set ProjectIsWritable to false if it is locked. Then I made that test code assert that ProjectIsWritable;

tombogle:
___
*[GlyssenShared/IProjectPersistenceReader.cs, line 10 at r3](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2KC2VQ7fOd9ZOc6HS2:-M2nP47O5x8mz4eHvJHj:b5zfez4) ([raw file](https://github.com/sillsdev/glyssen/blob/55c97d1a9f75dbc5b7584ba5c62abedf58f8e749/GlyssenShared/IProjectPersistenceReader.cs#L10)):*

> ```
> GetAllCustomReferenceTexts
> ```
Since this can have a predicate, and the All doesn't seem particularly helpful, anyway, should it be removed from the method name?
</blockquote></details>

I'm torn. The predicate is an "exclude" not a "where", so it suggests that the "normal" thing is to get them all. I think you're right that "All" is not strictly necessary, but it does kind of help to emphasize that we're getting all...except. Since it is only used in one particular situation (albeit the only instance in production code), it effectively functions as a bool. I'd like to make it a bool (and reverse its polarity to mean "even if already in the list of available one"), but the implementer has no access to the method that indicates that and it is an implementation detail that feels meaningless in the context of the interface.
I've tentatively made that method public (and tweaked it so it won't crash when s_allAvaialble is null) and changed the implementation of GetAllCustomReferenceTexts to unconditionally call that method to exclude any project that is already loaded. Seems to work okay. If we think that is a good idea, we'd need to rename it to something like GetCustomReferenceTextsNotAlreadyLoaded. That does kind of put the onus on the implementer to do that call correctly (though in one sense we were already requiring that indirectly). It's really primarily just an optimization to avoid getting the readers over and over, and that optimization may be less important in other implementations. I can just change AttemptToAddCustomReferenceText to dispose the reader if its for one that is already loaded.
How does that sound?

tombogle:
___
*[GlyssenShared/IProjectPersistenceWriter.cs, line 9 at r3](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2KCv2A4UlU3Fj6hf7V:-M2nZSQ_-P9ToOto7GUz:b-pkqz9d) ([raw file](https://github.com/sillsdev/glyssen/blob/55c97d1a9f75dbc5b7584ba5c62abedf58f8e749/GlyssenShared/IProjectPersistenceWriter.cs#L9)):*

> ```
> two
> ```
three?
</blockquote></details>

Deleted the comment. The tests now do in fact use these methods for IReferenceTextProject.

tombogle:
___
*[GlyssenShared/IProjectPersistenceWriter.cs, line 25 at r3](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2KDHVF8RWUM8Xu6Z2z:-M2n_DahC2X7DViERQ2x:b-994tcx) ([raw file](https://github.com/sillsdev/glyssen/blob/55c97d1a9f75dbc5b7584ba5c62abedf58f8e749/GlyssenShared/IProjectPersistenceWriter.cs#L25)):*
> ArchiveBookNoLongerAvailable
renamed to ArchiveBookThatIsNoLongerAvailable

tombogle:
___
*[GlyssenSharedTests/ReferenceTextTestUtils.cs, line 92 at r5](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2iztpH5-bb1M4FO5yU:-M2naZ17-MZmdBz0SN1E:b9j8nbu) ([raw file](https://github.com/sillsdev/glyssen/blob/a167cce397904121a02d7c070d968f696e0f33a1/GlyssenSharedTests/ReferenceTextTestUtils.cs#L92)):*

Is this method called? (It isn't in the stuff open in Reviewable currently.)
Regardless, I think it could be used in AddBook above to simplify the switch a bit by setting fileContents outside the switch.
</blockquote></details>

Yes, it's used in ReferenceTextTests.

tombogle:
___
*[GlyssenEngine/Project.cs, line 1111 at r5](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2eiYo04Yr_dtiKdvJM:-M2o6-ZcBo_GFvHgmlkh:byt3q36) ([raw file](https://github.com/sillsdev/glyssen/blob/a167cce397904121a02d7c070d968f696e0f33a1/GlyssenEngine/Project.cs#L1111)):*
> ArchiveBookNoLongerAvailable
ArchiveBookThatIsNoLongerAvailable

tombogle:
___
*[GlyssenEngineTests/ReferenceTextTests.cs, line 4207 at r5](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2iyElR64qTCd0Tpt_f:-M2oA3053COppUmE4AYH:b-dffohu) ([raw file](https://github.com/sillsdev/glyssen/blob/a167cce397904121a02d7c070d968f696e0f33a1/GlyssenEngineTests/ReferenceTextTests.cs#L4207)):*

Either of your suggestions makes sense. Perhaps more the latter if only one of three use the return value. (And I guess that's what I meant with my initial suggestion above to introduce a new method.)
</blockquote></details>

I did both

tombogle:
___
*[GlyssenFileBasedPersistenceTests/PersistenceImplementationTests.cs, line 68 at r5](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2nksZ-DY1CYp4MB7gm:-M2oB4P86107JYHQzXfx:b-54mjud) ([raw file](https://github.com/sillsdev/glyssen/blob/a167cce397904121a02d7c070d968f696e0f33a1/GlyssenFileBasedPersistenceTests/PersistenceImplementationTests.cs#L68)):*
> NoCustomReferenceTextsExist_GetsEmptyCollection
Changed to GetStandardReferenceText_NoCustomReferenceTextsExist_GetsReaderForMetadataOfRequestedReferenceText

tombogle:
___
*[GlyssenShared/IProjectPersistenceReader.cs, line 22 at r3](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2KCZr54VHx6IjnHMi-:-M2oCEc20XINNwTs4aIy:b7bl1tc) ([raw file](https://github.com/sillsdev/glyssen/blob/55c97d1a9f75dbc5b7584ba5c62abedf58f8e749/GlyssenShared/IProjectPersistenceReader.cs#L22)):*

This might be worth discussing. I think I get it after reading the comment a few times...
</blockquote></details>

Let's do that.

tombogle:
___
*[GlyssenSharedTests/ReferenceTextTestUtils.cs, line 92 at r5](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2iztpH5-bb1M4FO5yU:-M2tJ-GD2DiaBR6y2fNx:b-4zpy4l) ([raw file](https://github.com/sillsdev/glyssen/blob/a167cce397904121a02d7c070d968f696e0f33a1/GlyssenSharedTests/ReferenceTextTestUtils.cs#L92)):*

Did you see my suggestion to use it to simplify AddBook above?
</blockquote></details>

I debated doing that. I wasn't sure I wanted to have to do the switch twice, but I suppose it's not going to make much difference and probably simpler code (no duplication) is more important than a nanosecond of test time.

tombogle:
___
*[GlyssenSharedTests/ReferenceTextTestUtils.cs, line 92 at r5](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2iztpH5-bb1M4FO5yU:-M2tOEcI3pS2GeiGb0lo:bbr5sk2) ([raw file](https://github.com/sillsdev/glyssen/blob/a167cce397904121a02d7c070d968f696e0f33a1/GlyssenSharedTests/ReferenceTextTestUtils.cs#L92)):*

I debated doing that. I wasn't sure I wanted to have to do the switch twice, but I suppose it's not going to make much difference and probably simpler code (no duplication) is more important than a nanosecond of test time. 
</blockquote></details>

I figured out a way to get rid of the big switch altogether, so now the code is much simpler.

tombogle:
___
*[GlyssenShared/IProjectPersistenceReader.cs, line 62 at r9](https://reviewable.io/reviews/sillsdev/glyssen/659#-M3D0lqWCGK8kCfKyLw5:-M3D0lqWCGK8kCfKyLw6:b-8locl3) ([raw file](https://github.com/sillsdev/glyssen/blob/244a409f4adfa256fe13e7ebc0cfdde75024ebf4/GlyssenShared/IProjectPersistenceReader.cs#L62)):*
> ```Smalltalk
> 
> 		/// <summary>
> 		/// Enumerates the existing books in for the specified project, getting a
> ```

add: in canonical order

tombogle:
___
*[GlyssenShared/IProjectPersistenceReader.cs, line 68 at r9](https://reviewable.io/reviews/sillsdev/glyssen/659#-M3DB0s926KUNqeWavSz:-M3DB0s926KUNqeWavT-:b5rzsl7) ([raw file](https://github.com/sillsdev/glyssen/blob/244a409f4adfa256fe13e7ebc0cfdde75024ebf4/GlyssenShared/IProjectPersistenceReader.cs#L68)):*
> ```Smalltalk
> 		IEnumerable<ResourceReader<string>> GetExistingBooks(IProject project);
> 
> 		bool TryInstallFonts(IUserProject project, string fontFamily, IFontRepository fontRepository);
> ```

Need comment for this

tombogle:
___
*[GlyssenShared/IProjectPersistenceReader.cs, line 22 at r3](https://reviewable.io/reviews/sillsdev/glyssen/659#-M2KCZr54VHx6IjnHMi-:-M3S-OMx5oHG-hbg7uq4:bclgpwo) ([raw file](https://github.com/sillsdev/glyssen/blob/55c97d1a9f75dbc5b7584ba5c62abedf58f8e749/GlyssenShared/IProjectPersistenceReader.cs#L22)):*

If I remember our discussion, we said leave the name the same but improve the comment.
</blockquote></details>

I've taken a stab at re-wording the comment in both the interface and in the file-based implementation of the method. (The wording in each place differs.) See if it's any better.

tombogle:
___
*[GlyssenEngine/SampleProject.cs, line 44 at r12](https://reviewable.io/reviews/sillsdev/glyssen/659#-M3RsHNmCzMwSbyy8Yu5:-M3S3gFb7n9Lh51Kjpcv:b77i20j) ([raw file](https://github.com/sillsdev/glyssen/blob/42da8870d6dff263e6155d3b4b783338d67b3b22/GlyssenEngine/SampleProject.cs#L44)):*

Something seems off here. Maybe it is just because this is one-off for the sample, but why do we need to set the font family twice (once for the project and once for the metadata)?
</blockquote></details>

Maybe you just looked at the diff/code wrong. I don't see what you're talking about. I searched for "FontFamily = " and "kSampleFontFamily" in this file, and this is the only line where we're setting this. (The SampleProjectStub class also returns it, but that's a different class, and it's not really "setting" it.)

tombogle:
___
*[Glyssen/MainForm.cs, line 547 at r5](https://reviewable.io/reviews/sillsdev/glyssen/659#-M3RzrJE4juUztzDXyiY:-M3S4vk31JpEqaScnhK8:bkh4f4n) ([raw file](https://github.com/sillsdev/glyssen/blob/a167cce397904121a02d7c070d968f696e0f33a1/Glyssen/MainForm.cs#L547)):*

> ```
> if (GlyssenDblTextMetadata.GetRevisionOrChangesetId(projFilePath) == bundle.Metadata.RevisionOrChangesetId)
> ```
Is this check not needed anymore because we feel like we have moved past the point of needing it?
It feels like this logic got removed with your change.
</blockquote></details>

I think the overall logic is still intact. Here's what it's supposed to do (I think):
If there are other projects for the same "publication" (i.e., bundle ID), and the SAME revision, then we open the SelectExistingProjectDlg to see if the user wants to open an existing one or create a new one. If not, then most of the time that means there isn't an existing one, so we create a new one (with the default name) and open it. BUT (and all this logic is now in GetAvailableDefaultProjectFilePath), there's a chance there could be an existing project for a DIFFERENT revision, in which case we add a revision qualifier to the name to get a unique default. Finally, there is the insanely unlikely scenario where there might not be an existing valid project, but there could be a non-empty folder with the default name we came up with. If that happens, then GetAvailableDefaultProjectFilePath figures out what number to append to get an new folder.

tombogle:
___
*[GlyssenFileBasedPersistence/PersistenceImplementation.cs, line 476 at r12](https://reviewable.io/reviews/sillsdev/glyssen/659#-M3RvEKW9VNaOSoMES6C:-M3S6tav00oxi7qEaa9Z:b-48a379) ([raw file](https://github.com/sillsdev/glyssen/blob/42da8870d6dff263e6155d3b4b783338d67b3b22/GlyssenFileBasedPersistence/PersistenceImplementation.cs#L476)):*
> add a number to ensure a unique
If I change add to adds, will that make it clear?

tombogle:
___
*[GlyssenShared/ResourceReader.cs, line 6 at r9](https://reviewable.io/reviews/sillsdev/glyssen/659#-M3DDRCD5gvOcT4-y606:-M3S85Oj0WiX--3gYqyS:bb6un1j) ([raw file](https://github.com/sillsdev/glyssen/blob/244a409f4adfa256fe13e7ebc0cfdde75024ebf4/GlyssenShared/ResourceReader.cs#L6)):*

Since the only IdType I ended up needing was string, this class could be simplified. 
</blockquote></details>

Done

tombogle:
___
*[GlyssenFileBasedPersistence/PersistenceImplementation.cs, line 488 at r12](https://reviewable.io/reviews/sillsdev/glyssen/659#-M3RwSs43Pc8EOc3DklX:-M3S8LYEAYJL1KqPDPHp:b-ohgay3) ([raw file](https://github.com/sillsdev/glyssen/blob/42da8870d6dff263e6155d3b4b783338d67b3b22/GlyssenFileBasedPersistence/PersistenceImplementation.cs#L488)):*

I'm having a hard time following why the other checks are for non-empty folders, but here we care if the project file exists.
</blockquote></details>

In order for the project to be for a different revision, it would have to exist. If there is a non-empty folder that doesn't have a project in it, then it's a regular collision, but not one that requires us to append the revision number.

tombogle:
___
*[GlyssenFileBasedPersistence/PersistenceImplementation.cs, line 107 at r9](https://reviewable.io/reviews/sillsdev/glyssen/659#-M3C0GhP9Y2PGhGuIVvT:-M3SCh973RrkczO2T0oN:b-4fy4u4) ([raw file](https://github.com/sillsdev/glyssen/blob/244a409f4adfa256fe13e7ebc0cfdde75024ebf4/GlyssenFileBasedPersistence/PersistenceImplementation.cs#L107)):*

See my comments about the accuracy of the algorithm?
</blockquote></details>

Here's something similar: https://stackoverflow.com/questions/2818509/why-use-buffers-to-read-write-streams. I don;t think this is where I copied it from, though. My original implementation wasn't right and was crashing. In the case of this code on StackOverflow, they aren't checking that they read a full buffer. Maybe Stream.Read is different, but with TextReader.Read, it crashes if you try to do a subsequent read after reading the last little bit (that isn't a full buffer).

tombogle:
___
*[GlyssenFileBasedPersistence/PersistenceImplementation.cs, line 107 at r9](https://reviewable.io/reviews/sillsdev/glyssen/659#-M3C0GhP9Y2PGhGuIVvT:-M3SE60H7WsdDnZZ6pgz:b84bi4) ([raw file](https://github.com/sillsdev/glyssen/blob/244a409f4adfa256fe13e7ebc0cfdde75024ebf4/GlyssenFileBasedPersistence/PersistenceImplementation.cs#L107)):*

Here's something similar: https://stackoverflow.com/questions/2818509/why-use-buffers-to-read-write-streams. I don;t think this is where I copied it from, though. My original implementation wasn't right and was crashing. In the case of this code on StackOverflow, they aren't checking that they read a full buffer. Maybe Stream.Read is different, but with TextReader.Read, it crashes if you try to do a subsequent read after reading the last little bit (that isn't a full buffer).
</blockquote></details>

Here's another example with two different ways to end the looping: https://csharp.developreference.com/article/25610741/Fastest+way+to+delete+the+first+few+bytes+of+a+file
I don't know if one of those is preferable.

tombogle:
___
*[GlyssenFileBasedPersistence/PersistenceImplementation.cs, line 107 at r9](https://reviewable.io/reviews/sillsdev/glyssen/659#-M3C0GhP9Y2PGhGuIVvT:-M3SEo5l5jPvs7s16oTE:bsdn5kt) ([raw file](https://github.com/sillsdev/glyssen/blob/244a409f4adfa256fe13e7ebc0cfdde75024ebf4/GlyssenFileBasedPersistence/PersistenceImplementation.cs#L107)):*

Here's another example with two different ways to end the looping: https://csharp.developreference.com/article/25610741/Fastest+way+to+delete+the+first+few+bytes+of+a+file
I don't know if one of those is preferable.
</blockquote></details>

I was just about to add calls to Flush inside the loop to ensure we can handle an arbitrarily large amount of data. But I found this: https://docs.microsoft.com/en-us/dotnet/api/system.io.streamwriter.autoflush?view=netframework-4.8. So I think I can set Autoflush to false (probably the default, but weird that this is not clearly documented) and then call Flush at the end. (Apparently, this is not strictly necessary, since Dispose seems to handle this.)

PR #660, opened 2020-03-24
Instead of having errors in versification files ignored, alert the user

tombogle:
Previous behavior (accidentally, because of a change in SIL.Scripture) suppressed the error reporting and used the resulting versification, which might be bogus. Now we alert the user as appropriate.

Note: this is a pre-requisite to completion of https://github.com/sillsdev/Glyssen/pull/659

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/660)

### DIFF
--- a/ControlDataIntegrityTests/ControlDataIntegrityTests.csproj
+++ b/ControlDataIntegrityTests/ControlDataIntegrityTests.csproj
@@ -97,10 +97,10 @@
       <HintPath>..\packages\ParatextData.9.1.1-beta\lib\netstandard2.0\PtxUtils.dll</HintPath>
     </Reference>
     <Reference Include="SIL.Core, Version=8.0.0.0, Culture=neutral, PublicKeyToken=cab3c8c5232dfcf2, processorArchitecture=MSIL">
-      <HintPath>..\packages\SIL.Core.8.0.0-beta0077\lib\net461\SIL.Core.dll</HintPath>
+      <HintPath>..\packages\SIL.Core.8.0.0-beta0091\lib\net461\SIL.Core.dll</HintPath>
     </Reference>
     <Reference Include="SIL.Scripture, Version=8.0.0.0, Culture=neutral, PublicKeyToken=cab3c8c5232dfcf2, processorArchitecture=MSIL">
-      <HintPath>..\packages\SIL.Scripture.8.0.0-beta0037\lib\net461\SIL.Scripture.dll</HintPath>
+      <HintPath>..\packages\SIL.Scripture.8.0.0-beta0091\lib\net461\SIL.Scripture.dll</HintPath>
     </Reference>
     <Reference Include="SIL.WritingSystems, Version=8.0.0.0, Culture=neutral, PublicKeyToken=cab3c8c5232dfcf2, processorArchitecture=MSIL">
       <HintPath>..\packages\SIL.WritingSystems.8.0.0-beta0077\lib\net461\SIL.WritingSystems.dll</HintPath>
--- a/ControlDataIntegrityTests/packages.config
+++ b/ControlDataIntegrityTests/packages.config
@@ -21,8 +21,8 @@
   <package id="NUnit.Extension.TeamCityEventListener" version="1.0.7" targetFramework="net472" />
   <package id="NUnit.Extension.VSProjectLoader" version="3.8.0" targetFramework="net461" />
   <package id="ParatextData" version="9.1.1-beta" targetFramework="net472" />
-  <package id="SIL.Core" version="8.0.0-beta0077" targetFramework="net472" />
-  <package id="SIL.Scripture" version="8.0.0-beta0037" targetFramework="net472" />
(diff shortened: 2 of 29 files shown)


tombogle:
___
*[GlyssenEngine/Utilities/StringExtensions.cs, line 80 at r1](https://reviewable.io/reviews/sillsdev/glyssen/660#-M3CzVxK7zm-WQGhnmtE:-M3DHtiR7fYJSNy2Mjnw:b-z0a97p) ([raw file](https://github.com/sillsdev/glyssen/blob/ac9f637f44821ff8a2336adb7738fadf3e428865/GlyssenEngine/Utilities/StringExtensions.cs#L80)):*

> ```
> 			if (truncated.Length <= to)
> 				return truncated;
> ```
Seems you don't need this condition. Isn't it met by the following one?
</blockquote></details>

So it seems. I deleted it and tests still pass.

tombogle:
___
*[GlyssenEngine/Utilities/StringExtensions.cs, line 83 at r1](https://reviewable.io/reviews/sillsdev/glyssen/660#-M3Cziy0B76Ap0MyNR92:-M3DIAF9BkfAr_b1qN-p:b3ennfv) ([raw file](https://github.com/sillsdev/glyssen/blob/ac9f637f44821ff8a2336adb7738fadf3e428865/GlyssenEngine/Utilities/StringExtensions.cs#L83)):*

> ```
> to
> ```
Shouldn't this be `to + ellipses.length`?
I see that's not the way you wrote the tests, but it seems like what the caller would expect. If my assumption is wrong, and the caller expects their own string to be truncated to `to` with the addition of the ellipes, that probably warrants a comment (well, either way it does, I guess).
</blockquote></details>

I have no idea. I pretty much overdesigned this I think. My actual needs were quite simple and not very exacting.

tombogle:
___
*[GlyssenEngine/Utilities/GlyssenVersificationTable.cs, line 61 at r1](https://reviewable.io/reviews/sillsdev/glyssen/660#-M3DPMDA8pV-EmZwOQ3C:-M3DzjmCEit8_CxfRQo1:brkj80q) ([raw file](https://github.com/sillsdev/glyssen/blob/ac9f637f44821ff8a2336adb7738fadf3e428865/GlyssenEngine/Utilities/GlyssenVersificationTable.cs#L61)):*

Do we need to do anything to guarantee this has been called before any public methods can be used?
</blockquote></details>

I just made the constructor private, so the only way it can get created is by calling Initialize. In actual fact, neither the public property nor the public method I added would do anything harmful if called on a table that wasn't hooked up as the implementation, but they also wouldn't do anything helpful.

tombogle:
___
*[GlyssenEngine/Utilities/GlyssenVersificationTable.cs, line 97 at r1](https://reviewable.io/reviews/sillsdev/glyssen/660#-M3DP_egEXcGfSwevxu3:-M3E-ZGN7doS2CNhoy1H:bpwv6as) ([raw file](https://github.com/sillsdev/glyssen/blob/ac9f637f44821ff8a2336adb7738fadf3e428865/GlyssenEngine/Utilities/GlyssenVersificationTable.cs#L97)):*

Do these strings come from Paratext?
I suppose it isn't worth it to copy all their translations over right now...
</blockquote></details>

They sort of come from Paratext. I shortened most of them and made them a little less Paratexty. It is extremely unlikely that any Glyssen user will ever see any of these. I was able to sabotage some things to test the functionality, but in practice Paratext gripes pretty loudly, so they probably wouldn't get very far without fixing the problems. I seriously thought about just cleaning up the code and leaving the errors unreported.

tombogle:
What WAS happening is that the base implementation always throws the exception immediately, but Paratext substitutes its own implementation that always batches the errors up to be reported later. But since we were not setting a ProgressUtils.Implementation, the "report later" thing was never happening. So one change was to hook up that plumbing. But the problem with that is that in the case where we just want to report and rollback (assuming that is what we really want to do when a bundle has a bogus versification file), it doesn't work. And in the case of a versification file with a lot of errors, it's really annoying. We could (and maybe should) talk about why/if we really want/need to handle the problem in three different ways.

tombogle:
___
*[Glyssen/Utilities/ProgressUtilsImpl.cs, line 16 at r1](https://reviewable.io/reviews/sillsdev/glyssen/660#-M3DR2mw14_4QZZXyxVG:-M3E2-eY0_wUAlktATO-:bov4hki) ([raw file](https://github.com/sillsdev/glyssen/blob/ac9f637f44821ff8a2336adb7738fadf3e428865/Glyssen/Utilities/ProgressUtilsImpl.cs#L16)):*

I didn't review this carefully, making the assumption this was basically copied from elsewhere. If that is not correct, let me know.
</blockquote></details>

It is copied from Paratext, but I hacked out a lot fo the really complex stuff that handles all kinds of cases that I don't expect to occur in Glyssen. I also did a few Resharper-assisted tweaks. But, yeah, it probably doesn't need a fine tooth comb.

PR #661, opened 2020-03-27
Implement in-memory persistence for tests

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/661)


tombogle:
___
*[InMemoryTestPersistence/PersistenceImplementation.cs, line 161 at r1](https://reviewable.io/reviews/sillsdev/glyssen/661#-M3SjayCEZ0wCE-yb9A9:-M3khB-H006ZIQqV7CRk:b-4zbrqu) ([raw file](https://github.com/sillsdev/glyssen/blob/a90158f7241c85a4e18e623368abd85ed785a7b4/InMemoryTestPersistence/PersistenceImplementation.cs#L161)):*

> ```
> m_finalHome
> ```
The term "final home" isn't speaking to me.
Could you at least add a comment?
</blockquote></details>

No. That's gone :-)

tombogle:
___
*[InMemoryTestPersistence/ProjectKeyComparer.cs, line 48 at r1](https://reviewable.io/reviews/sillsdev/glyssen/661#-M3Sgb3pCY3D_O9ivy8q:-M3kheJaDj0HxDEOvQB8:bsn0q0h) ([raw file](https://github.com/sillsdev/glyssen/blob/a90158f7241c85a4e18e623368abd85ed785a7b4/InMemoryTestPersistence/ProjectKeyComparer.cs#L48)):*

I'm surprised to see this list of properties doesn't match the list above in Equals.
</blockquote></details>

That's because you're not yet used to doing code reviews for people who make mistakes!

tombogle:
___
*[GlyssenEngineTests/GlyssenEngineTests.csproj, line 326 at r2](https://reviewable.io/reviews/sillsdev/glyssen/661#-M3mevNIBQDxS0hSLxjc:-M3pys4108TX_cpuPcCk:b-f9nrf8) ([raw file](https://github.com/sillsdev/glyssen/blob/04f38e7e6fcf6b3b0071e646f3b8cb34c40e0b89/GlyssenEngineTests/GlyssenEngineTests.csproj#L326)):*

Uh. Was this intended?
</blockquote></details>

Yes. As the commit message says: "Added reference to GlyssenFileBasedPersistence in GlyssenEngineTests because it depends on InMemoryTestPersistence, which needs an instance of the file-based PersistenceImplementation to load the standard reference texts." Oddly, it built fine and tests ran fine (through Resharper) on my machine without it. But apparently without it, the DLL does not get shadow-copied by the normal NUnit test runner on the command line.
This is where in an ideal world we would have the reader and writer implementations split so that the GlyssenEngineTests could depend solely on a DLL with a class that implements the Reader logic. But probably not worth the effort and the overhead of another DLL.
Another possibility would be to implement methods in GlyssenEngine itself to load the standard reference texts from file and not use the persistence layer for them at all. However, I think maybe Vessel's plan is to deploy all reference texts via CouchBase.
Another possibility would be to (re)implement the necessary methods directly in the in-memory persistence implementation to load the standard reference texts from file. That would entail some code duplication, but maybe not a whole lot.

tombogle:
___
*[GlyssenEngineTests/ProjectTests.cs, line 348 at r2](https://reviewable.io/reviews/sillsdev/glyssen/661#-M3mlK198MvEH6_Uwe8K:-M3q0TcN9vGhvsnCRgeJ:b-3f3i6u) ([raw file](https://github.com/sillsdev/glyssen/blob/04f38e7e6fcf6b3b0071e646f3b8cb34c40e0b89/GlyssenEngineTests/ProjectTests.cs#L348)):*

> ```
> [Timeout(11000)]
> ```
Was the plan to take all these out?
</blockquote></details>

I considered it, but it appears that the file access was not the bottleneck. I'm going to do a little profiling to see if I can speed them up. I hate slow unit tests.

tombogle:
___
*[GlyssenEngineTests/ViewModelTests/AssignCharacterViewModelTests.cs, line 2179 at r2](https://reviewable.io/reviews/sillsdev/glyssen/661#-M3mjbxX1AR1tS6cN8LF:-M3rrPa3EeQrB0fB54g4:bnx289w) ([raw file](https://github.com/sillsdev/glyssen/blob/04f38e7e6fcf6b3b0071e646f3b8cb34c40e0b89/GlyssenEngineTests/ViewModelTests/AssignCharacterViewModelTests.cs#L2179)):*

I don't understand this addition. I guess you are just trying to get the test project back into a clean state?
Seems especially odd since the test doesn't reference m_testProject directly.
Comment?

Same below.
</blockquote></details>

CreateModel does reference m_testProject, and since methods are called on the model that modify the data, we need to put it back into pristine shape so it doesn't affect another test. In other fixtures we use a member variable to tell us that a full refresh is needed. That's probably better since we can avoid the overhead when just running a single test.

tombogle:
___
*[InMemoryTestPersistence/InMemoryTestPersistence.csproj, line 19 at r2](https://reviewable.io/reviews/sillsdev/glyssen/661#-M3mg9YH6SSJ-CctBIjZ:-M3rvAvH-YZiMJ8OjSJ8:b-4kvqqa) ([raw file](https://github.com/sillsdev/glyssen/blob/04f38e7e6fcf6b3b0071e646f3b8cb34c40e0b89/InMemoryTestPersistence/InMemoryTestPersistence.csproj#L19)):*

Hm. Definitely didn't expect this dependency, either...
</blockquote></details>

Same reason as the other one.

tombogle:
___
*[GlyssenEngineTests/GlyssenEngineTests.csproj, line 326 at r2](https://reviewable.io/reviews/sillsdev/glyssen/661#-M3mevNIBQDxS0hSLxjc:-M3rvKwa4varyyBx-ARr:b4fvhqz) ([raw file](https://github.com/sillsdev/glyssen/blob/04f38e7e6fcf6b3b0071e646f3b8cb34c40e0b89/GlyssenEngineTests/GlyssenEngineTests.csproj#L326)):*

Sorry; I missed that commit comment. There are so many, I don't see them anymore. (Are you planning to squash?)

It's just tests, so I don't think it is worth much effort to change.
</blockquote></details>

I don't usually squash. It seems to be more trouble than it's worth to me.

PR #662, opened 2020-05-02
PG-1329: Recognize "he said" text and automatically apply to other blocks

tombogle:
Initial implementation to detect and store reporting clauses and generate the corresponding "he said" reference blocks in GetBlocksForVerseMatchedToReferenceText.

This includes the migration of the "heSaid" element to "reportingClause" (now a list)
(Also fixed a few typos/misspellings)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/662)

### DIFF
--- a/GlyssenEngine/Bundle/GlyssenDblTextMetadata.cs
+++ b/GlyssenEngine/Bundle/GlyssenDblTextMetadata.cs
@@ -230,7 +230,7 @@ namespace GlyssenEngine.Bundle
 		}
 
 		/// <summary>
-		/// This is not part of the original DBL metadata. This data is now stored as part of the "langauge" data.
+		/// This is not part of the original DBL metadata. This data is now stored as part of the "language" data.
 		/// </summary>
 		[XmlElement("fontSizeInPoints")]
 		[DefaultValue(default(int))]
--- a/GlyssenEngine/Project.cs
+++ b/GlyssenEngine/Project.cs
@@ -526,6 +526,20 @@ namespace GlyssenEngine
 			return itemToRemove != null && ProjectCharacterDetail.Remove(itemToRemove);
 		}
 
+		public IReadOnlyList<string> ReportingClauses => m_metadata.Language.ReportingClauses;
+
+		public bool AddNewReportingClauses(IEnumerable<string> reportingClauses)
+		{
+			var newAdditions = reportingClauses.Where(c => !m_metadata.Language.ReportingClauses.Contains(c)).ToList();
+			if (newAdditions.Any())
+			{
+				m_metadata.Language.ReportingClauses.AddRange(newAdditions);
+				return true;
+			}
+
+			return false;
+		}
+
 		public void UpdateSettings(ProjectSettingsViewModel model, string defaultFontFamily, int defaultFontSizeInPoints, bool rightToLeftScript)
 		{
 			Debug.Assert(!IsNullOrEmpty(model.RecordingProjectName));
@@ -1384,7 +1398,7 @@ namespace GlyssenEngine
 				var sourceBookScript = sourceProject.m_books.SingleOrDefault(b => b.BookId == targetBookScript.BookId);
 				if (sourceBookScript != null)
 				{
-					targetBookScript.ApplyUserDecisions(sourceBookScript, ReferenceText);
+					targetBookScript.ApplyUserDecisions(sourceBookScript, ReferenceText, ReportingClauses);
 				}
 			}
 			Analyze();
--- a/GlyssenEngine/Project.cs
+++ b/GlyssenEngine/Project.cs
@@ -530,14 +530,10 @@ namespace GlyssenEngine
 
 		public bool AddNewReportingClauses(IEnumerable<string> reportingClauses)
 		{
-			var newAdditions = reportingClauses.Where(c => !m_metadata.Language.ReportingClauses.Contains(c)).ToList();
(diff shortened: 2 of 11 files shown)


tombogle:
___
*[GlyssenEngine/Project.cs, line 529 at r1](https://reviewable.io/reviews/sillsdev/glyssen/662#-M6fIzFV3F81etcPDA0g:-M78xStJ1Yc_iWssUQUD:b78xict) ([raw file](https://github.com/sillsdev/glyssen/blob/3643df3ae0dae1c27bea6e924577410582476594/GlyssenEngine/Project.cs#L529)):*

Any reason to have this be a list rather than a set? 
Set would simplify the add method below, anyway.

Later:
I see you are using the ordering to make the first one the default. Does it help at all to simply add a new field defining the default? Or does that unnecessarily complicate other things?
</blockquote></details>

I don't know. I wrestled with this and went back and forth. Since we already had a separate field with the "default", I initially considered adding a set to hold other ones, but that felt clunky and of course complicated the logic when looking for matches. The current approach is "okay" -- performance when adding is not critical and the number of items in the list will typically be small -- but it's a little yucky to have an ordered list when all we really need to know is which one is the default.

tombogle:
___
*[GlyssenEngine/Project.cs, line 529 at r1](https://reviewable.io/reviews/sillsdev/glyssen/662#-M6fIzFV3F81etcPDA0g:-M790I3Z6utdY8zL3psE:b4zpfth) ([raw file](https://github.com/sillsdev/glyssen/blob/3643df3ae0dae1c27bea6e924577410582476594/GlyssenEngine/Project.cs#L529)):*

Ok, well, probably not worth changing at this point.
</blockquote></details>

I guess the other option is to keep the "he said" text as a completely different thing conceptually. In a reference text, we would have a single he said member and no (other) reporting clauses (or if they were present, they would simply be ignored). In regular projects, we would no "he said" member (or at least no way to set it programmatically) and just have a list of reporting clauses. Maybe that would make more sense since we currently really have no basis for saying that the first one is the default.

tombogle:
___
*[GlyssenEngine/Project.cs, line 529 at r3](https://reviewable.io/reviews/sillsdev/glyssen/662#-M7DfAav-GGM7eqTX-6L:-M7DhzKr2hRardNF5KJj:bls2wjl) ([raw file](https://github.com/sillsdev/glyssen/blob/d9a9ac88bb8aefff97a31010ad5ce8797e415bc8/GlyssenEngine/Project.cs#L529)):*

Is this a collection simply because we don't have a readonly set?
</blockquote></details>

Yes. I'm assuming that the actual underlying implementation will be able to take advantage of the Set's fast Contains method.

tombogle:
___
*[GlyssenShared/Bundle/GlyssenDblMetadataLanguage.cs, line 52 at r3](https://reviewable.io/reviews/sillsdev/glyssen/662#-M7Dg6yV-Dp93c2GhCPW:-M7DiHPvDwgCX7-yLf7G:b2li7w2) ([raw file](https://github.com/sillsdev/glyssen/blob/d9a9ac88bb8aefff97a31010ad5ce8797e415bc8/GlyssenShared/Bundle/GlyssenDblMetadataLanguage.cs#L52)):*

I'm slightly confused. I thought you decided to have HeSaidText only apply to reference texts and ReportingClauses only apply to vernacular projects.

Beyond that, I don't see any problem with what you've done.
</blockquote></details>

That's understandable. In practice, this won't have any effect since currently only reference texts have heSaid set and since they never access their ReportingClauses collection.  So other than taking up an additional couple bytes of memory, this does nothing. I guess it's kind of my way of documenting that a "he said" really is a reporting clause, in case that ever matters in the future.

tombogle:
___
*[GlyssenShared/Bundle/GlyssenDblMetadataLanguage.cs, line 64 at r3](https://reviewable.io/reviews/sillsdev/glyssen/662#-M7DfRTk4sVyrxHeYmbx:-M7Dkh370WfdLgs5XZ_M:b580my7) ([raw file](https://github.com/sillsdev/glyssen/blob/362544949032b1f30fd9226a49ee3b614212a7a2/GlyssenShared/Bundle/GlyssenDblMetadataLanguage.cs#L64)):*

Since this is a set now, I think there are several places where you no longer need to check if the set contains an element then add it. You can just add it. This includes a few lines above here and Project::AddNewReportingClauses.
</blockquote></details>

In AddNewReportingClauses, I do need to know whether any new ones were added because the return value depends on that. I guess I could just compare the before/after counts or call Add in a loop (because that does report whether additions were made.

tombogle:
___
*[GlyssenShared/Bundle/GlyssenDblMetadataLanguage.cs, line 64 at r3](https://reviewable.io/reviews/sillsdev/glyssen/662#-M7DfRTk4sVyrxHeYmbx:-M7DlWkPAXUEUo6nyODt:b-b2jnqr) ([raw file](https://github.com/sillsdev/glyssen/blob/362544949032b1f30fd9226a49ee3b614212a7a2/GlyssenShared/Bundle/GlyssenDblMetadataLanguage.cs#L64)):*

In AddNewReportingClauses, I do need to know whether any new ones were added because the return value depends on that. I guess I could just compare the before/after counts or call Add in a loop (because that does report whether additions were made.
</blockquote></details>

Or I could do this: return reportingClauses.Aggregate(false, (current, reportingClause) => current | m_metadata.Language.ReportingClauses.Add(reportingClause)); // A bit obtuse and ugly?

PR #663, opened 2020-05-12
PG-1347/PG-1349: Prevent exception when reapplying matchups

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/663)


tombogle:
Not yet, but if/when I see the data, I might be able to come up with something better.

PR #670, opened 2020-05-18
PG-1356: Added steps to sign the engine

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/670)

### DIFF
--- a/build/Glyssen.proj
+++ b/build/Glyssen.proj
@@ -15,105 +15,98 @@
 		<RestartBuild Condition="Exists('$(RootDir)/packages/SIL.BuildTasks/tools/SIL.BuildTasks.dll')">false</RestartBuild>
 	</PropertyGroup>
 
-  <UsingTask TaskName="StampAssemblies"
-    AssemblyFile="$(RootDir)/packages/SIL.BuildTasks/tools/SIL.BuildTasks.dll"
-    Condition="Exists('$(RootDir)/packages/SIL.BuildTasks/tools/SIL.BuildTasks.dll')" />
-  <UsingTask TaskName="MakeWixForDirTree"
-    AssemblyFile="$(RootDir)/packages/SIL.BuildTasks/tools/SIL.BuildTasks.dll"
-    Condition="Exists('$(RootDir)/packages/SIL.BuildTasks/tools/SIL.BuildTasks.dll')" />
-  <UsingTask TaskName="Split"
-    AssemblyFile="$(RootDir)/packages/SIL.BuildTasks/tools/SIL.BuildTasks.dll"
-    Condition="Exists('$(RootDir)/packages/SIL.BuildTasks/tools/SIL.BuildTasks.dll')" />
-  <UsingTask TaskName="FileUpdate"
-    AssemblyFile="$(RootDir)/packages/SIL.BuildTasks/tools/SIL.BuildTasks.dll"
-    Condition="Exists('$(RootDir)/packages/SIL.BuildTasks/tools/SIL.BuildTasks.dll')" />
-  <UsingTask TaskName="NUnit3" AssemblyFile="$(RootDir)/packages/SIL.BuildTasks/tools/SIL.BuildTasks.dll" />
-  <Import Project="../.nuget/NuGet.targets" />
+	<UsingTask TaskName="StampAssemblies"
+		AssemblyFile="$(RootDir)/packages/SIL.BuildTasks/tools/SIL.BuildTasks.dll"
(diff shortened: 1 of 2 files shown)


build/Glyssen.proj
```diff
@@ -59,11 +59,11 @@
 	<CreateProperty Value="$(Version).$(BuildHash)">
 		<Output PropertyName="FullVersion" TaskParameter="Value"/>
 	</CreateProperty>
-	<CreateProperty Value="GlyssenInstaller.$(Version).exe" Condition=" '$(Label)' == '' ">
-		<Output PropertyName="InstallerFileName" TaskParameter="Value"/>
+	<CreateProperty Value="$(RootDir)\output\installer\GlyssenInstaller.$(Version).exe" Condition=" '$(Label)' == '' ">
+		<Output PropertyName="InstallerPathWithVersion" TaskParameter="Value"/>
```
ermshiperete, line 63: This is ok. However, since touching these lines it might be better to refactor them: msbuild now allows to use `PropertyGroup` inside of targets, so there's no longer a need to use `CreateProperty`. IMO this would make it easier to read.

```
<PropertyGroup>
    <FullVersion>$(Version).$(BuildHash)</FullVersion>
    etc.
</PropertyGroup>
```

build/Glyssen.proj
```diff
@@ -211,12 +211,18 @@
 	depending on whether your project's output MUST be signed or it's only nice if it is. For example, a continuous build
 	might use SignIfPossible, while a Publication build must use SignInstaller.-->
 	<Target Name="SignInstaller" DependsOnTargets="VersionNumbers; Installer">
-		<Exec Command='sign "$(RootDir)\output\installer\$(InstallerFileName)"'></Exec>
-		<Message Text="Signed installer at $(RootDir)\output\installer\$(InstallerFileName)" Importance="high"/>
+		<Exec Command='insignia -ib "$(InstallerPathWithVersion)" -o engine.exe'></Exec>
```
ermshiperete, line 214: I'm not familiar with insignia - I guess it gets installed as part of WIX?

build/Glyssen.proj
```diff
@@ -211,12 +211,18 @@
...
+		<Exec Command='sign engine.exe'></Exec>
```
ermshiperete, line 215: Or `<Exec Command='sign engine.exe'/>`

build/Glyssen.proj
```diff
@@ -211,12 +211,18 @@
...
+		<Exec Command='insignia -ab engine.exe "$(InstallerPathWithVersion)" -o $(InstallerPathWithVersion)'></Exec>
+		<Exec Command='sign "$(InstallerPathWithVersion)"'></Exec>
```
ermshiperete, line 217: Did you see https://wixtoolset.org/documentation/manual/v3/overview/insignia.html? That suggests that you could do it in the .wixproj file by overriding two targets there.

build/Glyssen.proj
tombogle: Yes, I think so. In the HearThis build log, this command results in the output: Windows Installer XML Toolset Inscriber version 3.11.2.4516...

tombogle:
PackageSources gets set in ../.nuget/NuGet.targets, which is imported on line 31.

ermshiperete:
> PackageSources gets set in ../.nuget/NuGet.targets, which is imported on line 31.

Maybe I'm missing something: on my machine NuGet.targets gets included, and it sets PackageSources, but it sets it to `PackageSource` which doesn't seem to be initialized (and when I run msbuild it fails because it can't find the nuget packages because we tell it to get the packages from '')

build/Glyssen.proj
tombogle: Done

build/Glyssen.proj
tombogle: Done

tombogle:
Based on the comment at line 20, I'm guessing it uses C:\Users\bogle\AppData\Roaming\NuGet to determine the sources.

build/Glyssen.proj
tombogle: I looked at it. Seems promising, but I'm a little sketchy on the details. Specifically, I don't know what our "sign" command does. My guess is that it's set up on the build agent to call **SignTool sign** with the appropriate command-line parameters. Also, I think that would then require some different plumbing to get ContinueOnError set correctly. I did take a stab at it, but I can't even get it to attempt to sign. I tried setting the **SignOutput** to true in both wixproj files (for the MSI and for the bootstrapper), but it doesn't seem to change anything so maybe I'm doing something wrong.

tombogle: See https://github.com/sillsdev/Glyssen/pull/671

PR #671, opened 2020-05-18
PG-1356: Non-critical follow-on task to set SignOutput true…

tombogle:
…and override certain WIX-defined targets to achieve signing.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/671)

### DIFF
--- a/build/Glyssen.proj
+++ b/build/Glyssen.proj
@@ -199,33 +199,6 @@
 		<Delete Files="$(RootDir)\output\RefTextUtilityInstallerinstaller\RefTextUtilityInstaller.msi" TreatErrorsAsWarnings="false" />
 
 	</Target>
-	
-	<!-- See https://wixtoolset.org/documentation/manual/v3/overview/insignia.html -->
-	<Target Name="SignCabs">
-		<PropertyGroup>
-			<IgnoreSigningErrors Condition="'$(IgnoreSigningErrors)' == ''">false</IgnoreSigningErrors>
-		</PropertyGroup>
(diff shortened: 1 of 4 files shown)


build/Glyssen.proj
```diff
@@ -199,23 +199,47 @@
 		<Delete Files="$(RootDir)\output\RefTextUtilityInstallerinstaller\RefTextUtilityInstaller.msi" TreatErrorsAsWarnings="false" />
 
 	</Target>
+	
+	<!-- See https://wixtoolset.org/documentation/manual/v3/overview/insignia.html -->
+	<Target Name="SignCabs">
```
ermshiperete, line 204: Does it work having the targets in this file? From the documentation I'd have thought that they should be in `Installer.wixproj`.

build/Glyssen.proj
```diff
@@ -199,23 +199,47 @@
...
+		<PropertyGroup>
```
ermshiperete, line 205: We could save some duplicated code by putting the property group outside of the target. I think it would still work.

build/Glyssen.proj
tombogle: I think you're right. It's been over a week now and I can only just barely remember what I was doing, but that's probably why I marked it WIP.

build/Glyssen.proj
tombogle: Now moved into wixproj files, so they still need to be duplicated in each file, but I have now moved them outside their build targets in the wixproj files. In a normal build, the property will be set correctly based on the builld target (SignInstaller vs. SignIfPossible), so the properties in the wixproj files are just needed as fallbacks in case the installer is built in some other context (e.g., Visual Studio).

PR #676, opened 2020-05-27
PG-1367: Corrected error-handling in quote parser and project loading

tombogle:
---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/676)

### DIFF
--- a/Glyssen/MainForm.cs
+++ b/Glyssen/MainForm.cs
@@ -440,6 +440,7 @@ namespace Glyssen
 
 		private void LoadProject(string filePath, Action additionalActionAfterSettingProject = null)
 		{
+			Logger.WriteEvent($"Attempting to load project from {filePath}");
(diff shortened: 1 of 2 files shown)


tombogle:
___
*[GlyssenEngine/Script/Block.cs, line 405 at r1](https://reviewable.io/reviews/sillsdev/glyssen/676#-M8Ly_9H8Q18LmVuNnWm:-M8QCzZQ-nmnBxG-azF6:biicy6k) ([raw file](https://github.com/sillsdev/glyssen/blob/1784c6adf5fe55e91341420edac05938515fc2ad/GlyssenEngine/Script/Block.cs#L405)):*

It's been drilled into my brain that I should never throw `Exception`. Would `ApplicationException` be better?

Also, is it worth a comment here explaining why you want to throw here? i.e. for debugging.
</blockquote></details>

It was just temporary for debugging. Gone now.

PR #678, opened 2020-05-29
PG-1373: Made the move down context menu item in the character or del…

tombogle:
…ivery column work correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/678)


tombogle:
Yeah, I kind of wondered about that too. My only concern/thought was that if we ever added another one (why would we), it might just work for free. But now I'm thinking that maybe two separate methods would be better. They could share the implementation that is common, which is the middle part of the method.

PR #683, opened 2020-06-08
PG-1337: Added warning when reference text does not cover all books in the project

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/683)


tombogle:
It's true. I discovered a mistake in the way I created the new control for PG-1376 when I was fixing the same mistake for PG-1337. Didn't quite seem worth a whole separate PR.

PR #686, opened 2020-06-10
Stripped off non-numeric portion of version number to prevent crash

tombogle:
when passing it to L10nSharp

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/686)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 
+## [1.3.11] - 2020-05-27
+
+### Added
+
+- When a block is aligned to "He said" in the reference text, Glyssen automatically scans other blocks to align any others with the same text.
+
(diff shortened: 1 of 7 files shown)


tombogle:
___
*[Glyssen/MainForm.cs, line 1110 at r2](https://reviewable.io/reviews/sillsdev/glyssen/686#-M9VXjxr9wTe9eHqTSEO:-M9sXB_5E7oNqPhsy7PP:b-r5gzsx) ([raw file](https://github.com/sillsdev/glyssen/blob/13b708f5bf650816998669b7a3ed3feaf2027520/Glyssen/MainForm.cs#L1110)):*

Just so you know, this will cause old data to have one value in analytics and new data to have another.
So comparison will be harder if you ever cared.
Likely it isn't an issue since you're just looking at things in mixpanel.
</blockquote></details>

Yes, I did consider that, but I figured that if we were looking at this we'd most likely be focused primarily on the question of how often this event occurred. if we needed to drill down into the details, it would be on a case-by-case basis.

PR #687, opened 2020-06-10
PG-1376: Made display of English version of reference text update

tombogle:
based on selected reference text, since it may be out of sync with latest English version.
Also fixed logic that selects the current reference text so that if user localizes on the fly, the same one is re-selected.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/687)

### DIFF
--- a/DistFiles/reference_texts/Russian/russian.glyssen
+++ b/DistFiles/reference_texts/Russian/russian.glyssen
@@ -14,10 +14,8 @@
   <identification>
     <name>Synodal Translation (1876)</name>
     <nameLocal />
-    <systemId type="gbc" />
-    <systemId type="tms" />
-    <systemId type="reap" />
-    <systemId type="biblica" />
+    <systemId type="fcbhOT"></systemId>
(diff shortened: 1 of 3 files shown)


tombogle:
___
*[Glyssen/Dialogs/ProjectSettingsDlg.cs, line 152 at r1](https://reviewable.io/reviews/sillsdev/glyssen/687#-M9VYdGD5Co9gU8DYK-t:-M9sF2Nl15RJZDKCcdt6:b4pkglx) ([raw file](https://github.com/sillsdev/glyssen/blob/e6489dc4282748ef560b6f1f294e193fe56537cf/Glyssen/Dialogs/ProjectSettingsDlg.cs#L152)):*

Preference, I guess, but I don't see any benefit in introducing this variable, especially way up here.
My guess is it is left over from a previous iteration of the code.
</blockquote></details>

Nope, it's required to deal with the (rare) case where the user does on-the-fly localization. When there is an existing selection that differs from the one that was set when we came into this dialog and we reset the data source, the selection gets lost.

PR #690, opened 2020-06-16
Improvements to English and Russian reference texts (v. 101)

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/690)


tombogle:
___
*[DistFiles/reference_texts/Russian/REV.xml, line 3001 at r2](https://reviewable.io/reviews/sillsdev/glyssen/690#-MEdjuW6D3lnT8BtLiKj:-MEe2pDb6xl0bbg-Pmq3:b-aob785) ([raw file](https://github.com/sillsdev/glyssen/blob/be24137f62a77b2afc2b12ac43d2882f8a413a13/DistFiles/reference_texts/Russian/REV.xml#L3001)):*

I was thinking we don't care about deliveries in the reference text...
But if we do, shouldn't the Russian have this, too?
</blockquote></details>

I think it's at least mostly true that we don't care. There is at least one place where the delivery of a reference block is used, but in the vast majority of cases, we can infer the delivery from the character. I went ahead and made the change because it can't hurt anything and it makes sense.

PR #691, opened 2020-06-17
PG-1380: Solved bugs related to making projects inactive

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/691)

### DIFF
--- a/Glyssen/Controls/ExistingProjectsList.cs
+++ b/Glyssen/Controls/ExistingProjectsList.cs
@@ -145,10 +145,18 @@ namespace Glyssen.Controls
 			{
 				try
 				{
-					using (var reader = new StreamReader(new FileStream(SelectedProject, FileMode.Open)))
+					// The file stream for the metadata needs to be closed BEFORE we call SetHiddenFlag because that
+					// will need to open the stream with a write lock to save the modified metadata.
+					GlyssenDblTextMetadata metadata;
+					using (var stream = new FileStream(SelectedProject, FileMode.Open))
 					{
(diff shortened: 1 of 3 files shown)


tombogle:
___
*[GlyssenFileBasedPersistence/ProjectRepository.cs, line 72 at r1](https://reviewable.io/reviews/sillsdev/glyssen/691#-MA2an39892gQO2aWbZl:-MA2eijH9U9nQG4eGWIU:b12c59a) ([raw file](https://github.com/sillsdev/glyssen/blob/d0c5e94c93a3e47780b28e85e9808fdbaf932a7a/GlyssenFileBasedPersistence/ProjectRepository.cs#L72)):*

This doesn't seem right. The project name is the file name without the extension, right.
</blockquote></details>

No, the directory structure is: language locale ID\DBL Id\Recording Project Name\language locale ID.glyssen

tombogle:
___
*[GlyssenFileBasedPersistence/ProjectRepository.cs, line 72 at r1](https://reviewable.io/reviews/sillsdev/glyssen/691#-MA2an39892gQO2aWbZl:-MA2lzvABADVXzo1KAfF:b-fktqjx) ([raw file](https://github.com/sillsdev/glyssen/blob/d0c5e94c93a3e47780b28e85e9808fdbaf932a7a/GlyssenFileBasedPersistence/ProjectRepository.cs#L72)):*

Ok, got it. So there was a previous bug which was returning the language locale ID... (scary we don't have testing here...)

But with the new code, now I'm confused why we call `GetFileName(folder)`. Isn't `folder` the desired result at that point?
</blockquote></details>

Blame Microsoft for not creating a GetInnermostFolder method that does the same thing as GetFilename. As it's name fails to imply, GetDirectoryName, gets the whole directory path, not the name of the innermost folder containing the file.

PR #696, opened 2020-06-24
PG-1374: Made Reference text matching logic fully account for versification differences.

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/696)

### DIFF
--- a/GlyssenEngine/ReferenceText.cs
+++ b/GlyssenEngine/ReferenceText.cs
@@ -822,6 +822,18 @@ namespace GlyssenEngine
 			if (!verseSplitLocations.Any())
 				return false;
 
+			// Since blocks cannot cross chapter breaks, if a split location is at the start of a chapter, typically
+			// both texts will already have a break there. However, if there is a versification difference, the chapter
+			// break location in one may not be the chapter break location in the other. But since verse 0 does not
+			// usually map to anything, changing the versification does not push it back across the chapter divide.
+			// In that case, we FIRST need to back up one verse, THEN change the versification so that if the text
+			// being broken has these two verses combined in a single block, we'll detect this as a valid break location.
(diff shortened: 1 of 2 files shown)


tombogle:
___
*[GlyssenEngine/ReferenceText.cs, line 831 at r1](https://reviewable.io/reviews/sillsdev/glyssen/696#-MAcWXxI81ig0wdn3Zvy:-MAcilql2Cevk3JaH6o2:b-4457v6) ([raw file](https://github.com/sillsdev/glyssen/blob/8421c82986aa3d9d60d856c91cd310d635512649/GlyssenEngine/ReferenceText.cs#L831)):*

> ```
> 					if (loc.VerseNum == 0)
> 						loc.PreviousVerse();
> ```
I don't understand this logic. But clearly you added it because it was needed, so go ahead and merge. But you might want to consider a comment.
</blockquote></details>

In the process of writing the comment, it ocurred to me that it might not be correct in the case of Psalms, where verse 0 is used for the Hebrew subtitles in some versifications.

tombogle:
___
*[GlyssenEngine/ReferenceText.cs, line 831 at r1](https://reviewable.io/reviews/sillsdev/glyssen/696#-MAcWXxI81ig0wdn3Zvy:-MAcqDfI3lMadeq59Elp:b-34zhef) ([raw file](https://github.com/sillsdev/glyssen/blob/8421c82986aa3d9d60d856c91cd310d635512649/GlyssenEngine/ReferenceText.cs#L831)):*

In the process of writing the comment, it ocurred to me that it might not be correct in the case of Psalms, where verse 0 is used for the Hebrew subtitles in some versifications.
</blockquote></details>

I'm still going to push the code through as is for now because it seems to be a clear improvement, but I've added a REVIEW comment to follow up and make sure I didn't break anything in Psalms

PR #697, opened 2020-06-25
Upgraded ParatextData to 9.1.7-beta and all of the SIL DLLs to 8.0.0-beta0153

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/697)

### DIFF
--- a/ControlDataIntegrityTests/App.config
+++ b/ControlDataIntegrityTests/App.config
@@ -19,7 +19,7 @@
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="Microsoft.Extensions.DependencyModel" publicKeyToken="adb9793829ddae60" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-3.1.0.0" newVersion="3.1.0.0" />
+        <bindingRedirect oldVersion="0.0.0.0-3.1.5.0" newVersion="3.1.5.0" />
       </dependentAssembly>
       <dependentAssembly>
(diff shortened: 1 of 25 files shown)


tombogle:
___
*[GlyssenTests/GlyssenTests.csproj, line 78 at r1](https://reviewable.io/reviews/sillsdev/glyssen/697#-MAgijsx1D77qOFGBq21:-MAgronj4DLb4id8fzn4:b-896fix) ([raw file](https://github.com/sillsdev/glyssen/blob/7ae57bbbcba58851d8a5e2b071e2513d19c1fab8/GlyssenTests/GlyssenTests.csproj#L78)):*

> ```
> DialogAdapters.Gtk2.
> ```
This is only the test dll, so I guess if it would cause a problem it would show up in running the tests... but maybe it will cause a problem in the reference text utility?
But it seems like it should be a problem. Gtk2 is for Geckofx v45 and the unspecified Gtk3 is for Geckofx 60.
I think Glyssen and Paratext both use v60 now, right? But probably libpalaso is the one stuck back in v45 land?

Something definitely seems off because now you are importing two different DialogAdapters packages, but there is only one reference here. (In both packages, the .dll is named the same).

Anyway, this whole thing has caused much pain for us in Bloom. When the version of DialogAdapters wanted by libpalaso caused a mismatch, we were getting a very strange initialization error in L10nSharp when it was trying to generate Palaso.xlf in AppData. This doesn't always happen, so sometimes you have to delete the file to reproduce.
I think we've actually had to fix it multiple times now... in different ways. But at least one fix was to update app.config to map the version range correctly.

Let me know if you want to discuss.

</blockquote></details>

Done.

PR #704, opened 2020-07-07
PG-1386: Handle case a versification mapping splits a verse across a chapter break…

tombogle:
…where the reference text has the first two verses in the chapter in a single block.

Also, tweaked code that counts the number of completed "relevant" blocks so that a matchup group does not get counted as completed if it contains any blocks that do not have a valid character set.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/704)

### DIFF
--- a/GlyssenEngine/ReferenceText.cs
+++ b/GlyssenEngine/ReferenceText.cs
@@ -423,7 +423,8 @@ namespace GlyssenEngine
 						iRefBlock--;
 						continue;
 					default:
-						if (refInitStartVerse > vernInitStartVerse || currentVernBlock.MatchesReferenceText)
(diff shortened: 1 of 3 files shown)


tombogle:
___
*[GlyssenEngine/ViewModels/AssignCharacterViewModel.cs, line 223 at r1](https://reviewable.io/reviews/sillsdev/glyssen/704#-MBeQqJL4m0EGuGw5sPe:-MBefuip4CJKirsau2JK:b-8bqfym) ([raw file](https://github.com/sillsdev/glyssen/blob/426f52a2ffafe87eb66fcfa734973789d4f3afc0/GlyssenEngine/ViewModels/AssignCharacterViewModel.cs#L223)):*

I feel like I'm missing something.
The previous logic would increment this for each block. 
Now we do it at most once.
</blockquote></details>

The previous logic had a return after incrementing.

PR #710, opened 2020-07-28
PG-1394: Improved logic for applying reporting clauses when matching blocks

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/710)

### DIFF
--- a/GlyssenEngine/ReferenceText.cs
+++ b/GlyssenEngine/ReferenceText.cs
@@ -535,17 +535,6 @@ namespace GlyssenEngine
 							break;
 						}
 						vernBlockInVerseChunk.SetMatchedReferenceBlock(refBlockInVerseChunk);
-						if (indexOfVernVerseStart + i - 1 >= 0)
-						{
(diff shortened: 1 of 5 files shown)


tombogle:
___
*[GlyssenEngine/ReferenceText.cs, line 548 at r1](https://reviewable.io/reviews/sillsdev/glyssen/710#-MDLF4HHEWmv0a1V6hmY:-MDNEJad00pdU2Zgu9B7:b-u3wqf) ([raw file](https://github.com/sillsdev/glyssen/blob/ab747b5124b4f071a5110a8beb25672a5d57df4c/GlyssenEngine/ReferenceText.cs#L548)):*

Am I correct that this block is not related specifically to he saids? So should it have a corresponding non-he said test?

Could use a comment to explain what it is for.
</blockquote></details>

It really was for "he saids" but could theoretically be useful for some other weird edge case that I couldn't think of. I started documenting it, then I was going to refactor it (after improving the other similar code and realizing that this code was even more deficient), and finally I realized that some other changes I had made rendered this code obsolete, so I deleted it.

PR #714, opened 2020-08-13
PG-1396: Omit reporting clause block that does not match to target.

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/714)


tombogle:
___
*[DistFiles/ReleaseNotes.md, line 49 at r2](https://reviewable.io/reviews/sillsdev/glyssen/714#-MEjKLVm2OPyvLnJMnUM:-MEwecu67pYGXbgtitXk:bw38leb) ([raw file](https://github.com/sillsdev/glyssen/blob/b7ea90f7b64773f15e5ec41c6e92065a846b8e5c/DistFiles/ReleaseNotes.md#L49)):*

Just curious why the major bump now?
</blockquote></details>

I removed a public extension method that is part of the GlyssenEngine, so it's a breaking change. It's an extension method whose name and behavior are the same as one in libpalaso, so it's not needed. (I had to make a change because a new using (needed for a different extension method) resulting in an ambiguity. I debated leaving it in and using explicit namespaces in the code, but that kind of defeats the purpose of extension methods. But after all the dust has cleared, this new method is now only being used in tests, so I can probably handle it differently.

tombogle:
___
*[GlyssenEngine/ReferenceText.cs, line 157 at r2](https://reviewable.io/reviews/sillsdev/glyssen/714#-MEjML3S8EISkpx2FPnG:-MEwgZFm6ASXbxWRLPfv:bci7eke) ([raw file](https://github.com/sillsdev/glyssen/blob/b7ea90f7b64773f15e5ec41c6e92065a846b8e5c/GlyssenEngine/ReferenceText.cs#L157)):*

Sorry; I tried to do this on a Friday afternoon. Really, I did.
But it is not going to happen...
In fact, even on a Monday, I'm thinking I might need a walkthrough...
</blockquote></details>

If it helps, you can dump the regex into regex101.com and try it against some potential reporting clauses, such as "Jesus said to them," "He said," "They said unto the creepy frog prince," and "She said to him." I'm handling likely English pronouns separately to allow them to be lowercased (so we don't accidentally lowercase proper names) and I'm trying to weed out really complex reporting clauses (like the frog prince one) because they are more likely to need a user to really look at them to see what is going on.

tombogle:
___
*[GlyssenEngine/Utilities/StringExtensions.cs, line 68 at r1](https://reviewable.io/reviews/sillsdev/glyssen/714#-MEhmSMw7FOUv__WNUax:-MEwrph81y7NxH8Eq8Ka:b-g0lwgx) ([raw file](https://github.com/sillsdev/glyssen/blob/7c50659a28f94011a4b8de82b03f6c5afa841e83/GlyssenEngine/Utilities/StringExtensions.cs#L68)):*

> ```
> (\w|\s)+\w
> ```
This will trim leading punctuation, too. Is that desired?
Also, I'm not sure this will do what you want if the reporting clause is `he-said`.
Also, trim non word-forming is more accurate since it trims whitespace, but you use "punctuation" below instead.
Also, I'm wondering if you really do want to preserve leading whitespace in this case?
</blockquote></details>

Changed it. Got rid of this new method entirely.

tombogle:
___
*[GlyssenEngine/ReferenceText.cs, line 163 at r3](https://reviewable.io/reviews/sillsdev/glyssen/714#-MEwxEpY8Gna5qst1O1O:-MExW_zd3bDb4hql1khP:b-1kc8ja) ([raw file](https://github.com/sillsdev/glyssen/blob/0344aafcdfba886571c7ce5212074d8068ff6a2b/GlyssenEngine/ReferenceText.cs#L163)):*

This seems to rely on the assumption (which is currently true but may not always be) that English is always second if it a second exists, else it is first.
Don't we have access to which language this ref text is? Should we check that?
If, on the other hand, we have lots of logic around which makes this same assumption, it might not be worth it.
</blockquote></details>

Here's how theses properties are defined:
		public bool HasSecondaryReferenceText => m_referenceTextType != ReferenceTextType.English;
		public ReferenceText SecondaryReferenceText => GetStandardReferenceText(ReferenceTextType.English);
So, I think the assumption is pretty much etched in code. What would be kind of nice would be to be able to ask a block to give us it's English reference block. But blocks don't know what language they are in. (We could do it by just having the block return its reference block recursively until the reference block was null, which would break the strict assumption that there would never be more than two levels of reference text, but it would still assume that English is the deepest level.)

PR #716, opened 2020-08-15
For Vessel: Ignore Hypothetical quotes types when reference text = null

AndrewChristensenFCBH:
Ignore Hypothetical quotes types when the reference text is unavailable. This was done because Vessel does not assign a reference text.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/716)

### DIFF
--- a/GlyssenEngine/Quote/ParserCharacterRepository.cs
+++ b/GlyssenEngine/Quote/ParserCharacterRepository.cs
@@ -45,25 +45,30 @@ namespace GlyssenEngine.Quote
 				"narrator overrides. These should be applied only after user has done disambiguation.");
 			var result = m_cvInfo.GetCharacters(bookId, chapter, verses, versification,
 				includeAlternatesAndRareQuotes, includeNarratorOverrides);
-			var hypotheticalsToReplace = result.Where(cv =>
-				{
-					if (cv.QuoteType == QuoteType.Hypothetical)
-					{
-						var refTextBook = m_referenceText.GetBook(bookId);
-						if (refTextBook != null)
-							return !refTextBook.GetBlocksForVerse(chapter,
-								verses.First().StartVerse, verses.Last().EndVerse).Any(b => b.CharacterId == cv.Character);
-						// REVIEW: Should we replace hypotheticals if there is no reference text for this book?
-					}
-					return false;
-				}).Select(c => c.Character).ToList();
-			if (hypotheticalsToReplace.Any())
+			// m_referenceText will never be null for Glyssen users, so the test for null will not effect Glyssen.
+			// At this time, Vessel users never have a reference text, so this test prevents an exception being thrown.
+			// This logic may be changed in the future so that hypothetical speakers can be better handled. 
+			if (m_referenceText != null)
 			{
-				result.Add(new CharacterSpeakingMode(CharacterVerseData.GetStandardCharacterId(BCVRef.NumberToBookCode(bookId), CharacterVerseData.StandardCharacter.Narrator),
-					String.Empty, String.Empty, false, QuoteType.Quotation));
-				result.RemoveAll(e => e.QuoteType == QuoteType.Hypothetical && hypotheticalsToReplace.Contains(e.Character));
+				var hypotheticalsToReplace = result.Where(cv =>
+					{
+						if (cv.QuoteType == QuoteType.Hypothetical)
+						{
+							var refTextBook = m_referenceText.GetBook(bookId);
+							if (refTextBook != null)
+								return !refTextBook.GetBlocksForVerse(chapter,
+									verses.First().StartVerse, verses.Last().EndVerse).Any(b => b.CharacterId == cv.Character);
+							// REVIEW: Should we replace hypotheticals if there is no reference text for this book?
+						}
+						return false;
+					}).Select(c => c.Character).ToList();
+				if (hypotheticalsToReplace.Any())
+				{
+					result.Add(new CharacterSpeakingMode(CharacterVerseData.GetStandardCharacterId(BCVRef.NumberToBookCode(bookId), CharacterVerseData.StandardCharacter.Narrator),
+						String.Empty, String.Empty, false, QuoteType.Quotation));
+					result.RemoveAll(e => e.QuoteType == QuoteType.Hypothetical && hypotheticalsToReplace.Contains(e.Character));
+				}
 			}
-
 			return result;
 		}
 
--- a/GlyssenEngineTests/GlyssenEngineTests.csproj
+++ b/GlyssenEngineTests/GlyssenEngineTests.csproj
@@ -276,6 +276,7 @@
       <DependentUpon>Resources.resx</DependentUpon>
     </Compile>
     <Compile Include="Quote\NewTestamentBooksFirstComparerTests.cs" />
+    <Compile Include="Quote\ParserCharacterRepositoryTests.cs" />
     <Compile Include="Quote\QuoteParserTests.cs" />
     <Compile Include="Quote\QuoteSystemGuesserTests.cs" />
     <Compile Include="Quote\QuoteSystemTests.cs" />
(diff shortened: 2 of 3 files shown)


GlyssenEngineTests/Quote/QuoteParserTests.cs
```diff
@@ -5370,6 +5370,21 @@ public void Parse_MultipleInterruptionsInMiddleOfMultiBlockQuote_MultiBlockQuote
 			Assert.AreEqual(CharacterVerse.kScriptureCharacter, output[5].CharacterId);
 		}
 		#endregion
+		
+		[Test]
+		public void Parse_GetCharactersWhenReferenceTestIsMissing()
```
tombogle, line 5375: The name should be in the form <<method under test>>_<<condition>>_<<result>>. So for example:
GetCharacters_ReferenceTextIsNull_ReturnsOnly GetsOnlyCharacterForVerse()
GetCharacters_ReferenceTextIsNullAndVerseHasMultiplesCharacters_GetsAllCharactersForVerse()
Since the test for 19:10:6 and 40:7:8 (character and character2) are really the same condition, you could use [TestCase(19, 10, 6)] and [TestCase(40, 7, 8)] instead of [Test] to show that clearly.

andrew-polk:
___
*[GlyssenEngineTests/Quote/QuoteParserTests.cs, line 5375 at r1](https://reviewable.io/reviews/sillsdev/glyssen/716#-MEkLPqZypgtluqGyZF9-r1-5375:-MExcaAV4AMB6e1FsE3R:bf42hre) ([raw file](https://github.com/sillsdev/glyssen/blob/97c8373c6e9ae1d1fbf57488b53d60a009d71e79/GlyssenEngineTests/Quote/QuoteParserTests.cs#L5375)):*

The name should be in the form <<method under test>>_<<condition>>_<<result>>. So for example:
GetCharacters_ReferenceTextIsNull_ReturnsOnly GetsOnlyCharacterForVerse()
GetCharacters_ReferenceTextIsNullAndVerseHasMultiplesCharacters_GetsAllCharactersForVerse()
Since the test for 19:10:6 and 40:7:8 (character and character2) are really the same condition, you could use [TestCase(19, 10, 6)] and [TestCase(40, 7, 8)] instead of [Test] to show that clearly.
</blockquote></details>

@tombogle 
I think reviewable threw away something from your first line there.

GlyssenEngine/Quote/ParserCharacterRepository.cs
```diff
@@ -45,25 +45,30 @@ public ParserCharacterRepository(ICharacterVerseRepository cvInfo, ReferenceText
...
-						if (refTextBook != null)
-							return !refTextBook.GetBlocksForVerse(chapter,
-								verses.First().StartVerse, verses.Last().EndVerse).Any(b => b.CharacterId == cv.Character);
-						// REVIEW: Should we replace hypotheticals if there is no reference text for this book?
-					}
-					return false;
-				}).Select(c => c.Character).ToList();
-			if (hypotheticalsToReplace.Any())
+			// m_referenceText will never be null for Glyssen users, so the test for null will not effect Glyssen.
+			// At this time, Vessel users never have a reference text, so this test prevents an exception being thrown when the GetBook method is called.
```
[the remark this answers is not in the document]
AndrewChristensenFCBH: Changed comment as requested.

GlyssenEngineTests/Quote/QuoteParserTests.cs
AndrewChristensenFCBH: Changed test as requested.  There are now 2 tests.

GlyssenEngine/Quote/ParserCharacterRepository.cs
```diff
@@ -45,25 +45,30 @@ public ParserCharacterRepository(ICharacterVerseRepository cvInfo, ReferenceText
...
-						if (refTextBook != null)
-							return !refTextBook.GetBlocksForVerse(chapter,
-								verses.First().StartVerse, verses.Last().EndVerse).Any(b => b.CharacterId == cv.Character);
-						// REVIEW: Should we replace hypotheticals if there is no reference text for this book?
-					}
-					return false;
-				}).Select(c => c.Character).ToList();
-			if (hypotheticalsToReplace.Any())
+			// m_referenceText will never be null for Glyssen users, so the test for null will not effect Glyssen.
+			// At this time, Vessel users never have a reference text, so this test prevents an exception being thrown when the GlyssenEngine is called by Vessel.
```
tombogle, line 49: Just remove " when the GlyssenEngine is called by Vessel"

AndrewChristensenFCBH: OK sorry, done.

GlyssenEngineTests/Quote/QuoteParserTests.cs
```diff
@@ -5370,6 +5370,28 @@ public void Parse_MultipleInterruptionsInMiddleOfMultiBlockQuote_MultiBlockQuote
...
 		}
 		#endregion
+
+		[TestCase(19, 10, "6", "narrator-PSA")]
+		[TestCase(40, 7, "8", "Jesus")]
+		public void GetCharacters_ReferenceTextIsNull_ReturnsOnlyOneCharacterForVerse(int book, int chapter, string verse, string expectedCharacter)
+		{
+			var cvRepo = new ParserCharacterRepository(ControlCharacterVerseData.Singleton, null);
+
+			var character = cvRepo.GetCharacters(book, chapter, new Verse(verse));
```
tombogle, line 5380: In both this test and the next, the variable should be named "characters" since it represents a collection (even if there only ends up being one character in it).

GlyssenEngineTests/Quote/QuoteParserTests.cs
```diff
@@ -5370,6 +5370,28 @@ public void Parse_MultipleInterruptionsInMiddleOfMultiBlockQuote_MultiBlockQuote
...
+
+			Assert.AreEqual(expectedCharacter, character.First().ToString());
```
tombogle, line 5382: Since you're only expecting one character, use Single instead of First.

GlyssenEngineTests/Quote/QuoteParserTests.cs
```diff
@@ -5370,6 +5370,28 @@ public void Parse_MultipleInterruptionsInMiddleOfMultiBlockQuote_MultiBlockQuote
 			Assert.AreEqual(CharacterVerse.kScriptureCharacter, output[5].CharacterId);
 		}
 		#endregion
+
+		[TestCase(19, 10, "6", "narrator-PSA")]
```
tombogle, line 5374: Sorry I didn't catch this sooner, but these tests should be in a new fixture: ParserCharacterRepositoryTests. Otherwise, it looks as if you're testing QuoteParser.GetCharacters, and there is no such method. The new file should be added to the Quote folder (since the class you're testing is in GlyssenEngine\Quote).

GlyssenEngineTests/Quote/QuoteParserTests.cs
```diff
@@ -5370,6 +5370,28 @@ public void Parse_MultipleInterruptionsInMiddleOfMultiBlockQuote_MultiBlockQuote
...
+		}
+
+		[Test]
+		public void GetCharacters_ReferenceTextIsNull_ReturnsMultipleCharactersForVerse()
+		{
+			var cvRepo = new ParserCharacterRepository(ControlCharacterVerseData.Singleton, null);
+
+			var character = cvRepo.GetCharacters(45, 2, new Verse("15"));
+
+			Assert.AreEqual("Gentiles", character.First().ToString());
```
tombogle, line 5392: First and Last don't really make sense. GetCharacters returns a HashSet, so it is not really ordered. You're just getting lucky. You probably just want to check that the total number is as expected and that the set contains the known values. (FWIW, I found this a bit confusing because there really shouldn't be any characters in ROM 2:15, but the test data has them there, so I guess it's okay.)

AndrewChristensenFCBH: Once I moved them to their own class and file I was getting different values and you are right Romans 2:15 didn't have any characters.  Strange.

AndrewChristensenFCBH: See if my test is ok for multiple numbers.

PR #740, opened 2020-11-23
PG-1425: Overloaded GetSwappedReferenceText with logic to decide whether to move leading verse number…

tombogle:
…with text based on verse number position in the vernacular

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/740)

### DIFF
--- a/Glyssen/Dialogs/AssignCharacterDlg.cs
+++ b/Glyssen/Dialogs/AssignCharacterDlg.cs
@@ -1395,9 +1395,10 @@ namespace Glyssen.Dialogs
 			m_viewModel.GetRowIndicesForMovingReferenceText(down, currentRowIndex, out int iPreceding, out int iFollowing);
 			var rowA = m_dataGridReferenceText.Rows[iPreceding];
 			var rowB = m_dataGridReferenceText.Rows[iFollowing];
+			var vernBlocks = m_viewModel.CurrentReferenceTextMatchup.CorrelatedBlocks;
 			if (colPrimary.Visible)
-				SwapRefText(rowA, rowB, colPrimary.Index);
-			SwapRefText(rowA, rowB, colEnglish.Index);
+				SwapRefText(vernBlocks, currentRowIndex, rowA, rowB, colPrimary.Index);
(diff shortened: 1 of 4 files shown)


andrew-polk (changes requested):
> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/740)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[GlyssenEngine/Script/Block.cs, line 1223 at r1](https://reviewable.io/reviews/sillsdev/glyssen/740#-MNepsdJB4RntSJsUX_l:-MNepsdKBFx7vOnOpYz9:b-h9t4di) ([raw file](https://github.com/sillsdev/glyssen/blob/a8553fdaa78d2ef588c78129537e2baee974fef0/GlyssenEngine/Script/Block.cs#L1223)):*
> ```Smalltalk
> 		/// moved down.
> 		/// </summary>
> 		public static void GetSwappedReferenceText(string rowA, string rowB, out string newRowAValue, out string newRowBValue)
> ```

Why do we still need this? Should it be marked deprecated?
If it's just for unit tests, should we move this version into test code?

tombogle:
___
*[GlyssenEngine/Script/Block.cs, line 1223 at r1](https://reviewable.io/reviews/sillsdev/glyssen/740#-MNepsdJB4RntSJsUX_l:-MNfiYvJ2ZGsRDAei73M:btun3w2) ([raw file](https://github.com/sillsdev/glyssen/blob/a8553fdaa78d2ef588c78129537e2baee974fef0/GlyssenEngine/Script/Block.cs#L1223)):*

Why do we still need this? Should it be marked deprecated?
If it's just for unit tests, should we move this version into test code?
</blockquote></details>

No, I was just trying to avoid removing an existing public method, which would require a new major version (even though I know Vessel isn't using that method), but one of the other PRs already forced a breaking change, so I've gone ahead and removed the inferior version of this method.

andrew-polk (approved):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/740)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #743, opened 2020-11-24
Removed override of HandleCurrentBlockChanged from AssignCharacterViewModel

tombogle:
None of the tests need it, and I couldn't find any path through production code where it made any difference. (It's not doing any real harm, so it's low priority and deserves fairly careful scrutiny in code review.)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/743)


bryanwussow (approved):
👍

PR #745, opened 2020-11-25
PG-1321: In QuoteParser handle case where we begin to process a long …

tombogle:
…quote using the primary character (ignoring the Alternate) but then find that the quote continues on into a verse that we go back and use the Alternate instead.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/745)

### DIFF
--- a/GlyssenEngine/Project.cs
+++ b/GlyssenEngine/Project.cs
@@ -38,7 +38,7 @@ namespace GlyssenEngine
 {
 	public class Project : ProjectBase, IUserProject
 	{
-		public const int kParserVersion = 49;
+		public const int kParserVersion = 50;
 
 		private const double kUsxPercent = 0.25;
 		private const double kGuessPercent = 0.10;
--- a/GlyssenEngine/Quote/QuoteParser.cs
+++ b/GlyssenEngine/Quote/QuoteParser.cs
@@ -942,11 +942,26 @@ namespace GlyssenEngine.Quote
 							// slight chance the delivery could change. And an even slighter chance we could have two
 							// possible deliveries left after removing any other characters from this list. So we'll
 							// be conservative and just prune the list down by character.
-							characterSpeakingDetails.RemoveAll(cv => cv.Character != prevQuoteBlock.CharacterId);
-							Debug.Assert(characterSpeakingDetails.Any(),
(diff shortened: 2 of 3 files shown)


andrew-polk (changes requested):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/745)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[GlyssenEngine/Quote/QuoteParser.cs, line 954 at r1](https://reviewable.io/reviews/sillsdev/glyssen/745#-MNetZ6S0i2fbeREbF97:-MNetZ6TC_1GUDoaOU_Z:b-en789o) ([raw file](https://github.com/sillsdev/glyssen/blob/9ff3c77a397d50ec4a1886113f17bf5a40feafe0/GlyssenEngine/Quote/QuoteParser.cs#L954)):*
> ```Smalltalk
> 										earlierBlock.ChapterNumber, earlierBlock.AllVerses, m_versification, true).Any(cv =>
> 										cv.Character == newCv.Character))));
> 								Debug.Assert(characterSpeakingDetails.Any(),
> ```

Any reason we don't want this Debug.Assert for the else block anymore?

tombogle:
___
*[GlyssenEngine/Quote/QuoteParser.cs, line 954 at r1](https://reviewable.io/reviews/sillsdev/glyssen/745#-MNetZ6S0i2fbeREbF97:-MNhotTYASP1tC7N-F68:b-sw1sx6) ([raw file](https://github.com/sillsdev/glyssen/blob/9ff3c77a397d50ec4a1886113f17bf5a40feafe0/GlyssenEngine/Quote/QuoteParser.cs#L954)):*
> ```
> Debug.Assert(characterSpeakingDetails.Any(),
> ```
Yes, the new if statement about already checks to ensure that we will not be removing them all.  It wouldn't necessarily be bad to assert that we didn't remove them all.  With the existing logic, the only way characterSpeakingDetails could be empty after the call to RemoveAll would be if it were empty before that call. As I look at the overall logic, I think that is still possible if there were a missing entry in the CV control file, but I'm not sure we need the assertion to detect that case.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/745)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #792, opened 2021-08-05
Fixed problems in reference text matching (especially for single-voice books)

tombogle:
(similar to PG-1277, but a different scenario) when a verse is missing in the target text but it is combined with the start of the following verse in the reference text.
Also fixed some edge cases where the forceMatch flag was not being taken into account.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/792)

### DIFF
--- a/GlyssenEngine/ReferenceText.cs
+++ b/GlyssenEngine/ReferenceText.cs
@@ -523,7 +523,28 @@ namespace GlyssenEngine
 					}
 					else
 					{
-						currentVernBlock.SetUnmatchedReferenceBlocks(correspondingReferenceBlocks);
+						// Before blindly mismatching all the ref blocks, check to see if the first ref block contains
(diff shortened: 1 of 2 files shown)


andrew-polk (commented):
Can't say I understand well enough to check your logic, but I don't see any glaring issues.
If I need to check the logic, we probably need to discuss.

I guess you intend to add the Commits file?

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/792#-:-MgM1yTsBpsU-SqRcZnj:b-6xprci)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tombogle)

tombogle:
I can discuss if you want. I don't know what the Commits thing is. It only appears in Reviewable, not in Github or GitExtensions. Maybe it's some new thing in the way Reviewable displays the change.

PR #807, opened 2021-11-18
Moved "rooster" (Rare) into correct verse in MAT

tombogle:
Added a dev resource file to source control

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/807)


andrew-polk (approved):
Approving, but I didn't see the xlsx change since it is binary. Have you considered making that a text or csv file?

tombogle:
It's for Mark, so Excel seemed appropriate. Just wanted to put it somewhere where it wouldn't get lost in case I need to update it some more. Probably could have just put it in Drive, but since we keep other related Excel files in this folder, it seemed like an okay place.

PR #811, opened 2022-01-28
PG-1479: Omit Paratext projects from list if an exception is thrown creating proxy

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/811)

### DIFF
--- a/Glyssen.sln.DotSettings
+++ b/Glyssen.sln.DotSettings
@@ -44,6 +44,7 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Scripter/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=theophanies/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Unapplied/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Unloadable/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=USFM/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Versifications/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=versioning/@EntryIndexedValue">True</s:Boolean>
--- a/Glyssen/Controls/ExistingProjectsList.cs
+++ b/Glyssen/Controls/ExistingProjectsList.cs
@@ -85,18 +85,7 @@ namespace Glyssen.Controls
 					{
 						if (!existingProjects.Contains(scrText.Settings.DBLId))
 						{
-							IProjectInfo proxy;
-							try
-							{
-								proxy = new ParatextProjectProxy(scrText);
-							}
-							catch (PtxUtils.FileWriteException e)
-							{
-								ErrorReport.ReportNonFatalExceptionWithMessage(e,
-									LocalizationManager.GetString("Project.OmitUnloadableParatextProject",
-										"This project will be omitted."));
-								continue;
-							}
+							IProjectInfo proxy = new ParatextProjectProxy(scrText);
 							m_paratextProjectIds[proxy.Name] = scrText.Guid;
 							yield return new Tuple<string, IProjectInfo>(proxy.Name, proxy);
 						}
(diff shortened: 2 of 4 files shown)


FoolRunning (changes requested):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/811)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[Glyssen/Controls/ExistingProjectsList.cs, line 93 at r1](https://reviewable.io/reviews/sillsdev/glyssen/811#-MuW_tCW-Dh4p1zryxoR:-MuW_tCW-Dh4p1zryxoS:b-x0hjez) ([raw file](https://github.com/sillsdev/glyssen/blob/47aa5eb636ce9186b69522139390330e55963190/Glyssen/Controls/ExistingProjectsList.cs#L93)):*
> ```Smalltalk
> 								proxy = new ParatextProjectProxy(scrText);
> 							}
> 							catch (PtxUtils.FileWriteException e)
> ```

Since it's basically a call to a 3rd-party application you can't control (Paratext), don't you want to catch any exception, not just FileWriteExceptions?

tombogle:
___
*[Glyssen/Controls/ExistingProjectsList.cs, line 93 at r1](https://reviewable.io/reviews/sillsdev/glyssen/811#-MuW_tCW-Dh4p1zryxoR:-MuX1_pD77iNoQy8QGjC:bviy26y) ([raw file](https://github.com/sillsdev/glyssen/blob/47aa5eb636ce9186b69522139390330e55963190/Glyssen/Controls/ExistingProjectsList.cs#L93)):*
<details><summary><i>Previously, FoolRunning (Tim Steenwyk) wrote…</i></summary><blockquote>

Since it's basically a call to a 3rd-party application you can't control (Paratext), don't you want to catch any exception, not just FileWriteExceptions?
</blockquote></details>

Yeah, I debated that. A part of me wanted to be really conservative and limit the catch to the specific errors I know can happen. The constructor for ParatextProjectProxy itself (or the DblMetadataLanguage object it uses) could *theoretically* throw an exception is and then I'd want to know so I could analyze the problem and try to fix it. Of course, there's also the possibility of some kind of system/installation problem (wrong DLL version, etc,) that we wouldn't necessarily want to present to the user as a problem with the project (especially since it would then occur repeatedly for every single project). It'd probably be nice if there were a base exception type that could wrap all "expected" exceptions from Paratext. Once thing I could do would be to wrap the specific calls to Paratext properties more tightly, but that kind of feels like overkill since currently everything is simple property assignments with nothing interesting going on. A final approach would be to catch and batch up the failures. If any project succeeded and their were multiple failures of the same type, I could report the failures together and tell the user how many projects were affected. If they all failed, then a different message would probably be in order. Still probably overkill for a situation that has only happened once, and we don't know why.

tombogle:
___
*[Glyssen/Controls/ExistingProjectsList.cs, line 93 at r1](https://reviewable.io/reviews/sillsdev/glyssen/811#-MuW_tCW-Dh4p1zryxoR:-MuXD3rI3YhVk0JgM11_:brg66cd) ([raw file](https://github.com/sillsdev/glyssen/blob/47aa5eb636ce9186b69522139390330e55963190/Glyssen/Controls/ExistingProjectsList.cs#L93)):*

Yeah, I debated that. A part of me wanted to be really conservative and limit the catch to the specific errors I know can happen. The constructor for ParatextProjectProxy itself (or the DblMetadataLanguage object it uses) could *theoretically* throw an exception is and then I'd want to know so I could analyze the problem and try to fix it. Of course, there's also the possibility of some kind of system/installation problem (wrong DLL version, etc,) that we wouldn't necessarily want to present to the user as a problem with the project (especially since it would then occur repeatedly for every single project). It'd probably be nice if there were a base exception type that could wrap all "expected" exceptions from Paratext. Once thing I could do would be to wrap the specific calls to Paratext properties more tightly, but that kind of feels like overkill since currently everything is simple property assignments with nothing interesting going on. A final approach would be to catch and batch up the failures. If any project succeeded and their were multiple failures of the same type, I could report the failures together and tell the user how many projects were affected. If they all failed, then a different message would probably be in order. Still probably overkill for a situation that has only happened once, and we don't know why.
</blockquote></details>

I think I've come up with a better approach...

FoolRunning (commented):
:lgtm: 

> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/811#-:-MuXJnft7kweriK0QH_6:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @FoolRunning)

FoolRunning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/811#-:-MuXJrE5CTZh7uD6dUK8:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @FoolRunning)

PR #812, opened 2022-02-04
Upgraded to GitVersion 5.8.1 (from deprecated GitVersionTask)

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/812)


ermshiperete (approved):

ermshiperete (approved):
I think `<IsPackable>false</IsPackable>` only works with the new sdk-style csproj files where the Pack target is defined in one of the implicit includes.

PR #820, opened 2022-04-25
Corrected flaw in the to determine whether to send an analytics tracking event

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/820)

### DIFF
--- a/Glyssen/Dialogs/AssignCharacterDlg.cs
+++ b/Glyssen/Dialogs/AssignCharacterDlg.cs
@@ -10,7 +10,6 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Drawing;
-using System.Globalization;
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Windows.Forms;
@@ -25,6 +24,7 @@ using SIL.Extensions;
 using SIL.Reporting;
 using SIL.Scripture;
 using SIL.Windows.Forms.Extensions;
+using static System.Globalization.CultureInfo;
 using static System.String;
 using Analytics = DesktopAnalytics.Analytics;
 using AssignCharacterViewModel = GlyssenEngine.ViewModels.AssignCharacterViewModel<System.Drawing.Font>;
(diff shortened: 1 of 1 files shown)


andrew-polk (changes requested):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/820)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @tombogle)

___
*[`Glyssen/Dialogs/AssignCharacterDlg.cs` line 238 at r1](https://reviewable.io/reviews/sillsdev/glyssen/820#-N0WafRXBA8s9ZWPM4bp:-N0WafRXBA8s9ZWPM4bq:b-l2yk5w) ([raw file](https://github.com/sillsdev/glyssen/blob/fa905b92c8be05f5a4792f1d87f19da87f89de39/Glyssen/Dialogs/AssignCharacterDlg.cs#L238)):*
> ```Smalltalk
> 					{ "lastVerse", block.LastVerseNum.ToString(InvariantCulture) },
> 					{ "character", character.CharacterId }
> 				});
> ```

Not sure what drove the change exactly, but would it be worth adding another prop to distinguish between the cases?

tombogle:
*[`Glyssen/Dialogs/AssignCharacterDlg.cs` line 238 at r1](https://reviewable.io/reviews/sillsdev/glyssen/820#-N0WafRXBA8s9ZWPM4bp:-N0abg-g3FlgJsOa7XJM:b-2x3ymw) ([raw file](https://github.com/sillsdev/glyssen/blob/fa905b92c8be05f5a4792f1d87f19da87f89de39/Glyssen/Dialogs/AssignCharacterDlg.cs#L238)):*

Not sure what drove the change exactly, but would it be worth adding another prop to distinguish between the cases?
</blockquote></details>

Unless I made a mistake in looking at the code, I don't think there are two cases. IsQuote will always return true if UserConfirmed; therefore, !IsQuote will always be false. but !IsQuote (false) && userConfirmed (true) => false. If UserConfrimed is false, IsQuote may be either true or false, which means !IsQuote may be either false or true. But it doesn't matter since the larger expression will always be false if !UserConfirmed. So as the code stood, this analytics event would never fire. (I checked MixPanel, and it never has.)

andrew-polk (approved):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/820)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`Glyssen/Dialogs/AssignCharacterDlg.cs` line 238 at r1](https://reviewable.io/reviews/sillsdev/glyssen/820#-N0WafRXBA8s9ZWPM4bp:-N0b8P8d7D0-Q-MNzK04:bcsw1ji) ([raw file](https://github.com/sillsdev/glyssen/blob/fa905b92c8be05f5a4792f1d87f19da87f89de39/Glyssen/Dialogs/AssignCharacterDlg.cs#L238)):*

Unless I made a mistake in looking at the code, I don't think there are two cases. IsQuote will always return true if UserConfirmed; therefore, !IsQuote will always be false. but !IsQuote (false) && userConfirmed (true) => false. If UserConfrimed is false, IsQuote may be either true or false, which means !IsQuote may be either false or true. But it doesn't matter since the larger expression will always be false if !UserConfirmed. So as the code stood, this analytics event would never fire. (I checked MixPanel, and it never has.)  
</blockquote></details>

Ok, thanks.
Since I think I still don't have a full grasp of it, my only remaining question is if this still serves its original purpose (which I think was to see if the user ever needed to override our auto choice of making something narrator??)
If yes, feel free to merge.

PR #838, opened 2022-09-30
+semver:major Split out character data files and related classes into separate DLL

tombogle:
A separate nuget package that does not depend on ParatextData is needed for inclusion in Paratext (see YouTrack issue PTX-22878).
Fixed build warnings about outdated binding redirects Removed character group template classes that have not been used for years Did some minor refactoring and fixed typos

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/glyssen/838)


jwickberg (approved):
Looks good - was wondering if some of the changes were done using a tool and which one if it was.

PR #852, opened 2023-03-13
PG-1493: Rolled back MarkDig.Signed to v. 0.30.4

tombogle:
This DLL is referenced explicitly only in tests and and Reference Text Utility, but that causes the referenced version to end up in the output folder, where it is picked up by the installer. ParatextData.DLL references the slightly older version, which causes a FileLoadException when doing string extraction.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/Glyssen/852)


andrew-polk (approved):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/Glyssen/852)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tombogle)

PR #855, opened 2023-05-09
PTX-23350: Added b marker to list of follow-on paragraph styles

tombogle:
This keeps Glyssen's logic in sync with Paratext (until that regex is actually available in ParatextDall.DLL anfd we can just use the "official" version).


andrew-polk (approved):

tombogle:
Note that the production code change here is now irrelevant, since the regex has moved into Paratext and IsFollowOnParagraphStyle now relies on the version of that regex from ParatextData. The added test case is perhaps still worth checking. This is blocked for now until the PTX issue is prioritized and the corresponding test failures in Paratext are evaluated.

